PR #9, opened 2019-04-25
reduce hardcoded dependencies

ltalirz:
used `pipdeptree` to reduce number of hardcoded dependencies in Pipfile.
The installed dependencies are almost unchanged (just backports-abc is
gone which is probably not needed).


ltalirz:
This should make things easier for us to maintain going forward.
Once you've checked that this looks fine, please make a new release.

yakutovicha:
Just made a release 19.4.3

PR #18, opened 2019-09-20
Fix ipython version

ltalirz:


ltalirz:
the `futures` package that is contained in the latest release breaks circus
https://github.com/aiidateam/aiida-core/issues/3333
, i.e. we should make a new release of the aiidalab metapackage.

PR #27, opened 2020-02-03
Add app.py and config.py from aiidalab-home app.

yakutovicha:


csadorf:
I'd prefer to have a separate branch for this particular process, something like `revise-app-management` or so.

yakutovicha:
I am just extracting things from aiidalab-home app here. I would propose that your work will go on top of it.

csadorf:
> I am just extracting things from aiidalab-home app here. I would propose that your work will go on top of it.

Sure, but I'll immediately branch then.

yakutovicha:
wouldn't it makes sense to merge it to develop first? This way you can focus on the app mananger and I will do ther other necessary updates.

csadorf:
Yes, I think that's what I mean? We merge this change into develop and then I'll branch to migrate the app manager.

I just don't think it would be a good idea to base any other code on this for now, which is why I originally suggested to not merge into develop.

yakutovicha:
Ok. I thought the word 'immediately' in that contest means to branch out from the *current* branch

yakutovicha:
@csadorf the PR is ready.

csadorf:
@yakutovicha What exactly would you like me to look for in the review? I anticipate that there will be a lot of changes needed to support the use case that we are targeting, but that's not something I should check here, correct? In my understanding this PR and the related aiidalab-home PR#30 are just moving code, correct?

yakutovicha:
> @yakutovicha What exactly would you like me to look for in the review? I anticipate that there will be a lot of changes needed to support the use case that we are targeting, but that's not something I should check here, correct? In my understanding this PR and the related aiidalab-home PR#30 are just moving code, correct?

That is all correct.  Along with moving the code, I restructured it a little bit differently. I also created some classes, put some function into `utils.py`, etc.

That would be great, if you could check whether you agree with class names and, more generally, the API.

This is totaly on a voluntary basis. I can also just merge the PR.

aiidalab/app.py
```diff
@@ -0,0 +1,647 @@
...
+import requests
+import ipywidgets as ipw
+from dulwich.repo import Repo
+
+
+class AppNotInstalledException(Exception):
+    pass
+
+
+class Version(ipw.VBox):
```
csadorf:
```suggestion
class VersionSelector(ipw.VBox):
```
I think all widgets should have class names that clearly indicate that this is a widget.

aiidalab/app.py
```diff
@@ -0,0 +1,647 @@
...
+            options={},
+            description='Select version',
+            disabled=False,
+            style={'description_width': 'initial'},
+        )
+        self.info = ipw.HTML('')
+        super().__init__([self.selected, self.change_btn, self.info])
+
+
+class AiidalabApp():  # pylint: disable=attribute-defined-outside-init,too-many-public-methods,too-many-instance-attributes
```
csadorf: The term "AiiDA lab" is not a single word, so this should probably be:
```suggestion
class AiidaLabApp():  # pylint: disable=attribute-defined-outside-init,too-many-public-methods,too-many-instance-attributes
```

aiidalab/app.py
```diff
@@ -0,0 +1,647 @@
...
+        if not self.has_git_repo():
+            return """<font color="#D8000C"><i class='fa fa-times-circle'></i> Not a Git Repo</font>"""
+        if not self._git_url:
+            return """<font color="#D8000C"><i class='fa fa-times-circle'></i> No remote URL</font>"""
+        if self.git_update_available():
+            return """<font color="#9F6000"><i class='fa fa-warning'></i> Update Available</font>"""
+        return """<font color="#270"><i class='fa fa-check'></i> Latest Version</font>"""
+
+    def render(self):
+        """"Display the app."""
```
csadorf: It's showing the app widget, not the app, right? I think that should be made clear in the name of this function and its doc-string. Maybe `render_start_widget` or something is more appropriate?

aiidalab/app.py
```diff
@@ -0,0 +1,647 @@
...
+        self.name = name
+
+    def in_category(self, category):
+        # One should test what happens if the category won't be defined.
+        return category in self.categories
+
+    def _get_appdir(self):
+        return path.join(self.aiidalab_apps, self.name)
+
+    def is_installed(self):
```
csadorf: Another function that will not make any sense any more after I have applied my proposed revisions. I propose to ignore this for now, I'll create a PR that introduces all of these changes.

aiidalab/app.py
```diff
@@ -0,0 +1,647 @@
...
+            style={'description_width': 'initial'},
+        )
+        self.info = ipw.HTML('')
+        super().__init__([self.selected, self.change_btn, self.info])
+
+
+class AiidalabApp():  # pylint: disable=attribute-defined-outside-init,too-many-public-methods,too-many-instance-attributes
+    """Class to manage AiiDA lab app."""
+
+    def __init__(self, name, app_data, aiidalab_apps):  #, custom_update=False):
```
csadorf: In my proposed revisions, the app will be identified by its path and nothing else. I assume we are going to merge this regardless and then I will apply the breaking change?

aiidalab/app.py
```diff
@@ -0,0 +1,647 @@
...
+        return category in self.categories
+
+    def _get_appdir(self):
+        return path.join(self.aiidalab_apps, self.name)
+
+    def is_installed(self):
+        """The app is installed if the corresponding folder is present."""
+        return path.isdir(self._get_appdir())
+
+    def has_git_repo(self):
```
csadorf: I am likely going to propose to use a distinct class to manage "git-installed" apps, but I am ok for leaving these functions here for now as long as it is clear that they are likely going to be removed.

aiidalab/app.py
```diff
@@ -0,0 +1,647 @@
...
+    def install_button(self):
+        """Button to install the app."""
+        if not hasattr(self, '_install_button'):
+            self._install_button = ipw.Button(description="install")
+            self._install_button.on_click(self._install_app)
+            self._refresh_install_button()
+        return self._install_button
+
+    def _install_app(self, _):
+        """Installing the app."""
```
csadorf: I get the impression that this is class is handling too many concerns. It is representing the application itself, but also a stand-in, because in the former case we assume that it is something that *is installed*, and can be rendered etc., and in the second case it is something that *can be installed*. So the latter would be better handled by some `AiidaLabAppStoreEntry` class or so.

I propose to merge _as is_ and I am going to split these various concerns in my proposed revisions.

aiidalab/app.py
```diff
@@ -0,0 +1,647 @@
...
+        """Refresh uninstall button."""
+        if self.is_installed():
+            self.uninstall_button.disabled = False
+            self.uninstall_button.button_style = 'danger'
+        else:
+            self.uninstall_button.disabled = True
+            self.uninstall_button.button_style = ''
+
+    @property
+    def version(self):
```
csadorf: I am not convinced that the apps themselves should have a version separate from the package they are installed with. Also, how does that relate to the git-revision?

aiidalab/app.py
```diff
@@ -0,0 +1,647 @@
...
+        """App's version."""
+        if not hasattr(self, '_version'):
+            self._version = Version()
+            self._version.change_btn.on_click(self._change_version)
+            self._refresh_version()
+        return self._version
+
+    @property
+    def refs_dict(self):
```
csadorf: This should probably be an internal function.

aiidalab/appstore.py
```diff
@@ -0,0 +1,150 @@
...
+
+import ipywidgets as ipw
+from IPython.display import display
+
+from .app import AiidalabApp
+from .config import AIIDALAB_APPS
+from .utils import load_app_registry
+
+
+class AiidalabAppStore(ipw.HBox):
```
csadorf:
```suggestion
class AiidaLabAppStore(ipw.HBox):
```

aiidalab/home.py
```diff
@@ -0,0 +1,125 @@
...
+    <a href=./start.ipynb?move_up={name} title="Move it up"><i class='fa fa-arrow-up' style='color:#337ab7;font-size:2em;' ></i></a>
+    <a href=./start.ipynb?move_down={name} title="Move it down"><i class='fa fa-arrow-down' style='color:#337ab7;font-size:2em;' ></i></a>
+    """.format(name=name),
+                       layout=layout)
+    btn_box.layout.margin = "50px 0px 0px 0px"
+
+    return btn_box
+
+
+class AiidalabHome:
```
csadorf:
```suggestion
class AiidaLabHome:
```

aiidalab/app.py
yakutovicha: so should we maybe change it to `VersionSelectorWidget` instead?

aiidalab/app.py
yakutovicha: One question here. It's a minor thing, but still.

Previously, we had quite a number of discussions on how to name the service, and we came up with 'AiiDA lab'. So lab should be lower case. I think the same should apply to the class names.

aiidalab/app.py
yakutovicha: ok

aiidalab/app.py
yakutovicha: I would say so.

aiidalab/app.py
yakutovicha: As far as I know, they are used only internally for the moment. So it is fine.

aiidalab/app.py
yakutovicha: We had an offline discussion and agreed that the classes should be named according to the widely adopted camel case notation.

aiidalab/app.py
yakutovicha: We had an offline discussion and agreed that the classes should be named according to the widely adopted camel case notation.

yakutovicha: done

aiidalab/app.py
yakutovicha: I agree. Some time ago I've even made an issue (#28) about the Home App design. Happy to rediscuss that with you.

aiidalab/app.py
yakutovicha: I agree.

aiidalab/appstore.py
yakutovicha: done

aiidalab/home.py
yakutovicha: done

aiidalab/app.py
yakutovicha: done

aiidalab/app.py
yakutovicha: I agree, but let's keep it like this for the moment. If you are going to take git-handling functionality away from the class - I see no point in renaming the function now.

aiidalab/app.py
yakutovicha: Renamed to `render_app_manager_widget`

PR #35, opened 2020-02-12
Update django dep to 1.11.28

CasperWA:
While `aiida-core==1.0.1` depends on `django==1.11.25`, the newest `develop` version of `aiida-core` already shows that it can support `django~=2.2` with no issue. So setting the dependency of `aiidalab` to `django==1.11.28` is not a big issue, but will indeed solve a vulnerability.

When we update to `aiida-core==1.1.0`, however, this will be updated in any case. So it is in some ways also a temporary dependency bump - again, to remove the vulnerability of earlier `django` version.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -47,7 +47,7 @@ cycler==0.10.0
 decorator==4.4.1
 deepdiff==3.3.0
 defusedxml==0.6.0
-django==1.11.25
+django==1.11.28
 dulwich==0.19.15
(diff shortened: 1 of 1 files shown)


yakutovicha:
I wouldn't merge this, because in the software stack here, the aiida version is 1.0.1. When moving to aiida 1.1.0 the update would happen automatically.

csadorf:
@yakutovicha Feel free to revert.

PR #36, opened 2020-02-12
Revert "Update django dep to 1.11.28"

yakutovicha:
Reverts aiidalab/aiidalab#35

I did not test `aiida-core==1.0.1` compatibility with `django==1.11.28`. If @CasperWA tested this, feel free to cancel the PR.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -47,7 +47,7 @@ cycler==0.10.0
 decorator==4.4.1
 deepdiff==3.3.0
 defusedxml==0.6.0
-django==1.11.28
+django==1.11.25
 dulwich==0.19.15
 elementpath==1.1.8
 entrypoints==0.3


CasperWA:
> Reverts #35
> 
> I did not test `aiida-core==1.0.1` compatibility with `django==1.11.28`. If @CasperWA tested this, feel free to cancel the PR.

As I wrote in #35:
> While `aiida-core==1.0.1` depends on `django==1.11.25`, the newest `develop` version of `aiida-core` already shows that it can support `django~=2.2` with no issue. So setting the dependency of `aiidalab` to `django==1.11.28` is not a big issue, but will indeed solve a vulnerability.

On the other hand, I don't think this repository should depend on `django` at all, since it is not a direct dependency, but rather a dependency of `aiida-core`.

CasperWA:
In the end, all of this is mute, since `aiida-core==1.1.0` has [_just_ been release](aiidateam/aiida-core#3754).

yakutovicha:
The direct dependencies of this package are defined in the [Pipfile](https://github.com/aiidalab/aiidalab/blob/develop/Pipfile). The file is then used to automatically build the complete aiidalab environment (specified in `requirements.txt`) and making sure that there are no dependency clashes.

yakutovicha:
> In the end, all of this is mute, since `aiida-core==1.1.0` has [_just_ been release](aiidateam/aiida-core#3754).

Ah, this is good to know, I can then proceed with updating the environment.

PR #40, opened 2020-02-16
Pipfile: move packages from [dev-packages] to [packages] section.

yakutovicha:
Since aiida-core and many aiida plugins has been stabilized, they were
moved from [dev-packages] to [packages] section of the Pipfle. The 
requirements file has been adapted accordingly.

### DIFF
--- a/Pipfile
+++ b/Pipfile
@@ -5,30 +5,30 @@ verify_ssl = true
 
 [dev-packages]
 #aiida-ase = "~=1.0.1"
-#aiida-castep = "~=1.0.0"
-aiida-codtools = "~=2.0.0"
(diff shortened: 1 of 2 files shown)


yakutovicha:
I think I figured out the difference between [dev-packages] and [packages] section.

To my understanding, the dependency tree resolution goes through the following steps:
1. Resolve dependencies within the [dev-packages] section
1. Resolve dependencies within the [packages] section.
1. Merge those dependencies and, in case of problems, give priority to the dependencies of the [packages] section.

According to this schema, it makes sense why `tornado` dependency, that was set for `aiida-core` to be lower than `<5` was not respected. This happened only because `aiida-core` was in the [dev-packages] section, while the other packages required `tornado` to be `>5`

PR #73, opened 2020-07-01
Add GitHub actions continuous-integration workflow for testing.

csadorf:
I think the test structure highlights that maybe, decoupling the `aiidalab` package from the home app was the wrong move after all. Regardless, I suggest that we move forward in this way for now and reconsider this structure at a later stage.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,7 +1,6 @@
 name: continuous-integration
 
-on:
-  [push]
+on: [push, pull_request]
 
 jobs:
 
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yml
```diff
@@ -0,0 +1,39 @@
+name: continuous-integration
+
+on:
+  [push]
```
CasperWA, line 4: (Same comment as at aiidalab-home) Is there a reason why we're not testing also on PRs?
```suggestion
on: [push, pull_request]
```

csadorf: No.

CasperWA:
I wonder why this change didn't start the action run for this PR?

PR #79, opened 2020-07-16
Add aiida-nanotech-empa package as a dependency

yakutovicha:


yakutovicha:
Going to close this PR since managing dependencies is added to aiidalab in #82

PR #80, opened 2020-07-16
Revise installation release line specification

csadorf:
This is not 100% debugged yet, but I'm creating this draft PR in case you want to do some testing.

### DIFF
--- a/aiidalab/app.py
+++ b/aiidalab/app.py
@@ -569,25 +569,11 @@ class AppManagerWidget(ipw.VBox):
             ipw.HBox([self.modifications_indicator, self.modifications_ignore]),
         ]
 
-        def formatted_version(version):
-            if version is None:
-                return '[not installed]'
-            if not version:
-                return '[n/a]'
-            if version.startswith('git:refs/heads/'):  # branch
-                return f"{version[len('git:refs/heads/'):]} (latest)"
-            if version.startswith('git:refs/tags/'):  # tag
-                return version[len('git:refs/tags/'):]
-            if version.startswith('git:'):  # commit
-                return version[4:4 + 8]
-            raise ValueError("Unknown version format: '{}'".format(version))
-
         self.version_selector = VersionSelectorWidget()
         ipw.dlink((self.app, 'available_versions'), (self.version_selector.install_version, 'options'),
-                  transform=lambda versions: [(formatted_version(version), version) for version in versions])
+                  transform=lambda versions: [(self._formatted_version(version), version) for version in versions])
         ipw.dlink((self.app, 'installed_version'), (self.version_selector.installed_version, 'value'),
-                  transform=formatted_version)
-        ipw.dlink((self.app, 'busy'), (self.version_selector, 'disabled'))
+                  transform=self._formatted_version)
         self.version_selector.layout.visibility = 'visible' if with_version_selector else 'hidden'
         self.version_selector.disabled = True
         self.version_selector.install_version.observe(self._refresh_widget_state, 'value')
@@ -597,6 +583,26 @@ class AppManagerWidget(ipw.VBox):
 
         self.app.observe(self._refresh_widget_state)
         self.app.refresh_async()  # init all widgets
+
+    @staticmethod
+    def _formatted_version(version):
+        """Format the unambigious version identifiee to a human-friendly representation."""
+        if version is None:  # not installed
+            return '[not installed]'
+
+        if not version:  # will be displayed during transition phases
+            return '[n/a]'
+
+        if version.startswith('git:refs/heads/'):  # branch
+            return f"{version[len('git:refs/heads/'):]} (latest)"
+
+        if version.startswith('git:refs/tags/'):  # tag
+            return version[len('git:refs/tags/'):]
+
+        if version.startswith('git:'):  # commit
+            return version[4:4 + 8]
+
+        raise ValueError("Unknown version format: '{}'".format(version))
 
     def _refresh_widget_state(self, _=None):
         """Refresh the widget to reflect the current state of the app."""
@@ -616,8 +622,9 @@ class AppManagerWidget(ipw.VBox):
                 tooltip = "Operation blocked due to local modifications."
 
             # Determine whether we can install, updated, and uninstall.
-            can_install = not self.app.is_installed() \
-                    or self.app.installed_version != self.version_selector.install_version.value
+            installed = self.app.is_installed()
+            can_switch = self.app.installed_version != self.version_selector.install_version.value
+            can_install = can_switch or not installed
             can_uninstall = self.app.is_installed()
             try:
                 can_update = self.app.updates_available and not can_install
@@ -629,7 +636,8 @@ class AppManagerWidget(ipw.VBox):
             self.install_button.button_style = 'info' if can_install else ''
             self.install_button.icon = '' if can_install and not modified else warn_or_ban_icon if can_install else ''
             self.install_button.tooltip = '' if can_install and not modified else tooltip if can_install else ''
-            self.install_button.description = 'Install'
+            self.install_button.description = 'Install' if not (installed and can_switch) \
+                    else f'Install ({self._formatted_version(self.version_selector.install_version.value)})'
 
             # Update the uninstall button state.
             self.uninstall_button.disabled = busy or blocked or not can_uninstall
@@ -648,6 +656,10 @@ class AppManagerWidget(ipw.VBox):
                     "circle-up" if can_update and not modified else warn_or_ban_icon if can_update else ""
                 self.update_button.button_style = 'success' if can_update else ''
                 self.update_button.tooltip = '' if can_update and not modified else tooltip if can_update else ''
+
+            # Update the version_selector widget state.
+            more_than_one_version = len(self.version_selector.install_version.options) > 1
+            self.version_selector.disabled = busy or blocked or not more_than_one_version
 
             # Indicate whether there are local modifications and present option for user override.
             self.modifications_indicator.value = \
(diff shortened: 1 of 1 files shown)


csadorf:
There are still a few minor kinks to be worked out, but for the most part this should be working now.

To test this, add entries to your registry that look like this:
```
  "apps": {
        "aiidalab-hello-world": {
                "git_url": "/home/aiida/local/aiidalab-hello-world",
                "meta_url": "https://raw.githubusercontent.com/aiidalab/aiidalab-hello-world/master/metadata.json",
                "categories": ["utility"]
        },
        "aiidalab-hello-world@test": {
                "git_url": "/home/aiida/local/aiidalab-hello-world@test",
                "meta_url": "https://raw.githubusercontent.com/aiidalab/aiidalab-hello-world/master/metadata.json",
                "categories": ["utility"]
        },
        "aiidalab-hello-world@v2": {
                "git_url": "/home/aiida/local/aiidalab-hello-world@v2",
                "meta_url": "https://raw.githubusercontent.com/aiidalab/aiidalab-hello-world/master/metadata.json",
                "categories": ["utility"]
        },
        "aiidalab-hello-world@bf402510b5b5c4ac6b90bcef6c0e470d3c713e88": {
                "git_url": "/home/aiida/local/aiidalab-hello-world@bf402510b5b5c4ac6b90bcef6c0e470d3c713e88",
                "meta_url": "https://raw.githubusercontent.com/aiidalab/aiidalab-hello-world/master/metadata.json",
                "categories": ["utility"]
        },
```
Please note that I used a local git repository for testing, but this could of course also be a regular remote git repository hosted on GitHub etc.

aiidalab/app.py
```diff
@@ -82,14 +82,131 @@ class AiidaLabApp(traitlets.HasTraits):
...
+            """
+            assert self.short_ref or self.commit
+
+            if self.commit:  # The release line is a commit.
+                assert self.commit.encode() in self._repo.object_store
+                yield self.commit.encode()
+            else:
+                ref = self._resolve_short_ref(self.short_ref)
+                if ref is None:
+                    raise ValueError(f"Unable to resolve '{self.short_ref}'. "
```
CasperWA, line 161: Casting as `repr()` does the same. A nice trick :)
```suggestion
                    raise ValueError(f"Unable to resolve {self.short_ref!r}. "
```

aiidalab/app.py
```diff
@@ -82,14 +82,131 @@ class AiidaLabApp(traitlets.HasTraits):
...
+                        ref: c for ref, c in self._repo.refs.as_dict().items() if ref.startswith('refs/tags'.encode())
+                    }
+                    tags_lookup = {c: ref for ref, c in all_tags.items()}
+                    tagged_commits_on_head = [
+                        c.commit.id for c in self._repo.get_walker(self._repo.refs[ref]) if c.commit.id in tags_lookup
+                    ]
+                    yield tags_lookup.get(ref_commit, ref)
+                    for commit in tagged_commits_on_head:
+                        if commit != ref_commit:
+                            yield tags_lookup[commit]
```
CasperWA, line 177: Is it supposed to yield the first one here?

aiidalab/app.py
```diff
@@ -539,16 +673,16 @@ def _check_modified_state(self):
         if blocked:
             raise RuntimeError("Unable to perform operation, there are local modifications to the app repository.")
 
-    def _install_version(self, _):
+    def _install_version(self, _=None):
```
CasperWA, line 676: I guess there are times when this will be called without supplying a non-used parameter then? Will it ever be called _with_ a non-used parameter?

aiidalab/app.py
```diff
@@ -82,14 +82,131 @@ class AiidaLabApp(traitlets.HasTraits):
...
+        """Utility class to operate on the release line of the app.
+
+        A release line is specified via the app url as the part behind the '@'.
+
+        A release line can be specified either as
+            a) a commit denoted by a hexadecimal number with either 20 or 40 digits, or
+            b) a short reference, which can be either a branch or a tag name.
+
+        A full ref is the ref as defined in the Git glossary, e.g., 'refs/heads/main'.
+        A full ref and a commit are a revision.
```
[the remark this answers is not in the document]
csadorf: Actually either of them define a revision.
```suggestion
        A revision is either a full ref or a commit.
```

aiidalab/app.py
```diff
@@ -82,14 +82,131 @@ class AiidaLabApp(traitlets.HasTraits):
...
+        """
+
+        def __init__(self, app, line):
+            self.app = app
+            self.line = line
+
+            match = re.fullmatch(r'(?P<commit>([0-9a-fA-F]{20}){1,2})|(?P<short_ref>.+)', line)
```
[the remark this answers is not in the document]
csadorf: This regex will match a commit defined as SHA1 with length 20 or 40 (those two lengths are valid SHA1s for git). Anything else will be interpreted as *short_ref* (not check is made as part of the regex whether it is a valid ref).

aiidalab/app.py
csadorf: You mean on line 174? Yes.

aiidalab/app.py
csadorf: Yes, it is sometimes called with an argument, however, we do not use that argument hence why it is named `_`.

aiidalab/app.py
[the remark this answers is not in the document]
csadorf: The exact format of a valid ref is defined as part of the [git-check-ref-format specification](https://mirrors.edge.kernel.org/pub/software/scm/git/docs/git-check-ref-format.html). This check is already implement in dulwich, however there is no need to check for valid refs at this point.

csadorf:
Still remaining item: Reliably determine whether updates are available.

aiidalab/app.py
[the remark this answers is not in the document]
CasperWA: > This regex will match a commit defined as SHA1 with length 20 or 40 (those two lengths are valid SHA1s for git). Anything else will be interpreted as _short_ref_ (no check is made as part of the regex whether it is a valid ref).

Didn't know that, thanks :)

> The exact format of a valid ref is defined as part of the [git-check-ref-format specification](https://mirrors.edge.kernel.org/pub/software/scm/git/docs/git-check-ref-format.html). This check is already implemented in dulwich, however there is no need to check for valid refs at this point.

Sure.

aiidalab/app.py
CasperWA: So the logic is that the first thing returned is the `ref_commit` and all subsequent returned objects are `tagged_commits_on_head`?

aiidalab/app.py
CasperWA: Sure. I just didn't think this function would be called both with and without a parameter.
Is there a way to avoid this or does it not really matter? I'm more thinking about the "why" of this. But maybe not so important for now...

aiidalab/app.py
csadorf: Yes, exactly.

aiidalab/app.py
csadorf: This kind of pattern actually appears in various places whenever we use a function stand-alone (without the argument), and register as a callback function with an ipywidget, that will call it with itself as first argument. I'm not sure if there is a better way to do this.

csadorf:
> Still remaining item: Reliably determine whether updates are available.

I've tested this and it seems to be working.

csadorf:
@CasperWA This is ready for final review and probably also ready for merge.

aiidalab/app.py
CasperWA: Yeah, it's those pescy ipywidgets callbacks. Just didn't find the place this function was used for it. I don't think there is another way. Just curious about the double-effect and why it wouldn't be used for just one or the other.

aiidalab/app.py
```diff
@@ -103,6 +113,126 @@ class AppRegistryData:
...
+            """Attempt to resolve the short-ref to a full ref.
+
+            For example, 'branch' would be resolved to 'refs/heads/branch'
+            if 'branch' is a local branch or 'refs/tags/branch' if it was
+            a tag.
+
+            This function returns None if the short-ref cannot be resolved
+            to a full reference.
+            """
+            # Check if short-ref is a head:
```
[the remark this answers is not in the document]
csadorf: I am trying to stick to the [official git nomenclature](https://git-scm.com/docs/gitglossary) as much as possible, in so far a *head* is technically not necessarily a "local branch", however I think we could put that into parentheses to provide some more guidance:
```suggestion
            # Check if short-ref is a head (e.g. local branch):
```
Would that be sufficient?

aiidalab/app.py
```diff
@@ -215,28 +356,33 @@ def _has_git_repo(self):
...
-            version = 'git:refs/heads/' + AIIDALAB_DEFAULT_GIT_BRANCH
+        assert self._release_line is not None
 
         with self._show_busy():
-            assert version.startswith('git:refs/heads/')
-            branch = re.sub(r'git:refs\/heads\/', '', version)
+            if version is None:
+                version = f'git:{self._release_line.line}'
+
+            if not re.fullmatch(r'git:((?P<commit>([0-9a-fA-F]{20}){1,2})|(?P<short_ref>.+))', version):
```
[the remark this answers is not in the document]
csadorf: Yes, but those code paths are technically independent. We require the release version to be a valid ref, however the `.install_app()` method has a `version` argument that could theoretically be whatever. Hence the second somewhat redundant check.

aiidalab/app.py
```diff
@@ -41,12 +43,12 @@ class VersionSelectorWidget(ipw.VBox):
     """Class to choose app's version."""
 
     disabled = traitlets.Bool()
-    available_versions = traitlets.Dict(traitlets.Unicode())
 
     def __init__(self, *args, **kwargs):
         style = {'description_width': '100px'}
-        self.release_line = ipw.Dropdown(
-            description='Release line',
+        self.install_version = ipw.Dropdown(
```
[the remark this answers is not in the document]
csadorf: Yes, we can totally rename that to `version_to_install` for clarity. 👍

aiidalab/app.py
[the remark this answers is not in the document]
CasperWA: Yeah this is fine. I was just following along with the method description 👍

PR #81, opened 2020-07-22
Implement AppRegistryData data class to hold app registry data.

csadorf:
This change set refactors the app registry data management to be stored within a dataclass. This makes it easier to validate that data and keep it self-consistent. Furthermore, it makes it easier to understand where that data is coming from and check whether it is available in downstream functions.

### DIFF
--- a/aiidalab/app.py
+++ b/aiidalab/app.py
@@ -9,7 +9,7 @@ from contextlib import contextmanager
 from time import sleep
 from threading import Thread
 from subprocess import check_output, STDOUT, CalledProcessError
-from dataclasses import dataclass, field
+from dataclasses import dataclass, field, asdict
 
 from typing import List, Dict
 
@@ -128,6 +128,11 @@ class AiidaLabApp(traitlets.HasTraits):
         self.path = os.path.join(aiidalab_apps_path, self.name)
         self.refresh_async()
         self._watch_repository()
+
+    def __repr__(self):
+        app_data_argument = None if self._registry_data is None else asdict(self._registry_data)
+        return f"AiidaLabApp(name={self.name!r}, app_data={app_data_argument!r}, " + \
+                f"aiidalab_apps_path={os.path.dirname(self.path)!r})"
 
     @traitlets.default('modified')
     def _default_modified(self):
--- a/aiidalab/app.py
+++ b/aiidalab/app.py
@@ -131,8 +131,8 @@ class AiidaLabApp(traitlets.HasTraits):
 
     def __repr__(self):
         app_data_argument = None if self._registry_data is None else asdict(self._registry_data)
-        return f"AiidaLabApp(name={self.name!r}, app_data={app_data_argument!r}, " + \
-                f"aiidalab_apps_path={os.path.dirname(self.path)!r})"
+        return (f"<AiidaLabApp(name={self.name!r}, app_data={app_data_argument!r}, "
+                f"aiidalab_apps_path={os.path.dirname(self.path)!r})>")
 
     @traitlets.default('modified')
     def _default_modified(self):
--- a/aiidalab/app.py
+++ b/aiidalab/app.py
@@ -97,9 +97,9 @@ class AiidaLabApp(traitlets.HasTraits):
         """Dataclass that contains the app data from the app registry."""
         git_url: str
         meta_url: str
-        metainfo: Dict[str, str]
         categories: List[str]
         groups: List[str]  # appears to be a duplicate of categories?
+        metainfo: Dict[str, str] = field(default_factory=dict)
         gitinfo: Dict[str, str] = field(default_factory=dict)
         hosted_on: str = None
 
@@ -302,7 +302,7 @@ class AiidaLabApp(traitlets.HasTraits):
                     return json.load(json_file)
             except IOError:
                 return dict()
-        elif self._registry_data is not None:
+        elif self._registry_data is not None and self._registry_data.metainfo:
             return dict(self._registry_data.metainfo)
         else:
             raise RuntimeError(
--- a/aiidalab/app.py
+++ b/aiidalab/app.py
@@ -338,7 +338,9 @@ class AiidaLabApp(traitlets.HasTraits):
     @property
     def url(self):
         """Provide explicit link to Git repository."""
-        return self._registry_data.git_url
+        if self._registry_data is not None:
+            return self._registry_data.git_url
+        return None
 
     @property
     def more(self):
--- a/aiidalab/app.py
+++ b/aiidalab/app.py
@@ -9,8 +9,10 @@ from contextlib import contextmanager
 from time import sleep
 from threading import Thread
 from subprocess import check_output, STDOUT, CalledProcessError
+from dataclasses import dataclass, field
 
-import requests
+from typing import List, Dict
+
 import traitlets
 import ipywidgets as ipw
 from dulwich.porcelain import fetch
@@ -90,6 +92,17 @@ class AiidaLabApp(traitlets.HasTraits):
     busy = traitlets.Bool(readonly=True)
     modified = traitlets.Bool(readonly=True, allow_none=True)
 
+    @dataclass
+    class AppRegistryData:
+        """Dataclass that contains the app data from the app registry."""
+        git_url: str
+        meta_url: str
+        metainfo: Dict[str, str]
+        categories: List[str]
+        groups: List[str]  # appears to be a duplicate of categories?
+        gitinfo: Dict[str, str] = field(default_factory=dict)
+        hosted_on: str = None
+
     class AppPathFileSystemEventHandler(FileSystemEventHandler):
         """Internal event handeler for app path file system events."""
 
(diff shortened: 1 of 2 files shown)


csadorf:
@CasperWA I'd prefer to merge this before #80 .

aiidalab/app.py
```diff
@@ -90,6 +92,17 @@ class AiidaLabApp(traitlets.HasTraits):
     busy = traitlets.Bool(readonly=True)
     modified = traitlets.Bool(readonly=True, allow_none=True)
 
+    @dataclass
+    class AppRegistryData:
+        """Dataclass that contains the app data from the app registry."""
+        git_url: str
+        meta_url: str
+        categories: List[str]
+        groups: List[str]  # appears to be a duplicate of categories?
```
csadorf, line 101: @CasperWA Do you know what's up with that?

CasperWA: I don't know what `groups` refer to. But it may be something @yakutovicha introduced, either for internal use or not.
The official keys for the registry can be found in the README of the [`aiidalab-registry`](https://github.com/aiidalab/aiidalab-registry) repository. Here only `categories` appears.

aiidalab/app.py
```diff
@@ -90,6 +92,17 @@ class AiidaLabApp(traitlets.HasTraits):
     busy = traitlets.Bool(readonly=True)
     modified = traitlets.Bool(readonly=True, allow_none=True)
 
+    @dataclass
+    class AppRegistryData:
```
CasperWA, line 96: It seems you're fond of the "have a 'package' (here the class `AiidaLabApp`) come with everything inside it may need" :) What I'm trying to say - is there a specific reason you're not putting these "internal" classes outside `AiidaLabApp`? Having a bit more modularity ...

aiidalab/app.py
```diff
@@ -124,6 +129,11 @@ def __init__(self, name, app_data, aiidalab_apps_path):
         self.refresh_async()
         self._watch_repository()
 
+    def __repr__(self):
+        app_data_argument = None if self._registry_data is None else asdict(self._registry_data)
+        return f"AiidaLabApp(name={self.name!r}, app_data={app_data_argument!r}, " + \
+                f"aiidalab_apps_path={os.path.dirname(self.path)!r})"
```
CasperWA, line 135: Two suggestions in one:
- Wrap the representation in the "standard" `<>`.
- Write the return string value without the use of string concatenation and `\`.
```suggestion
        return (
            f"<AiidaLabApp(name={self.name!r}, app_data={app_data_argument!r}, "
            f"aiidalab_apps_path={os.path.dirname(self.path)!r})>"
        )
```

aiidalab/app.py
```diff
@@ -297,13 +307,11 @@ def metadata(self):
...
             except IOError:
                 return dict()
-        elif self._meta_info is not None:
-            return dict(self._meta_info)
-        elif self._meta_url is None:
+        elif self._registry_data is not None and self._registry_data.metainfo:
+            return dict(self._registry_data.metainfo)
+        else:
             raise RuntimeError(
                 f"Requested app '{self.name}' is not installed and is also not registered on the app registry.")
```
CasperWA, line 314: Very good conceptual fix this one (moving the raise to the final `else`)! 💪

aiidalab/app.py
```diff
@@ -330,7 +338,9 @@ def title(self):
     @property
     def url(self):
         """Provide explicit link to Git repository."""
-        return self._git_url
+        if self._registry_data is not None:
+            return self._registry_data.git_url
+        return None
```
CasperWA, line 343: This could be done like this, but what you're doing might be clearer, i.e., more explicit?
```suggestion
        return getattr(self._registry_data, "git_url", None)
```

aiidalab/app.py
csadorf: So it seems that `groups` is the key to the `categories` definitions in the app registry file, whereas `categories` contains the Category name. I think we can keep both for now, but we should probably make sure to harmonize that at some point.

aiidalab/app.py
csadorf: The classes are tightly linked to the `AiidaLabApp` class and are not supposed to be used outside of that context, hence why I've made them inner classes. This also reduces the "API surface" overall. However, I think this is mostly a matter of taste and we could take a different approach, but I would prefer to just leave it as is unless you strongly object.

aiidalab/app.py
csadorf: Generally a fan of creating nice one-liners with `getattr`, however in this case this would not work, because `_registry_data` itself can be `None`, not `git_url`.

aiidalab/app.py
CasperWA: No strong objections. It's merely a combination of a sudden realization and stating it "out loud" to start a small talk :)

That said, I completely agree with your assessment here and reasoning behind it 👍

aiidalab/app.py
CasperWA: But wouldn't `getattr` then return `None` as that's set to be the fallback/default value? Or would it completely fail due to an inconsistent type of `self._registry_data`? I'll just try this.

CasperWA: Yup, in Python 3.7.5 I get that this will work fine, even if `self._registry_data` is `None`. It'll just return the third input parameter for `getattr`, i.e., `None`.

aiidalab/app.py
CasperWA: Right. So one is the "look up" and the other is the values for any given app? Or did I misunderstand?

aiidalab/app.py
csadorf: TIL! :)

aiidalab/app.py
csadorf: No, the categories are actually defined in the metadata registry as part of the top-level `categories` group. The `apps/<app-id>/groups` values match the `categories/<category-id>` keys. The `apps/<app-id>/categories` values actually contain the formatted name of the categories. This means that the categories of apps are over-specified.

csadorf:
@CasperWA This should be ready for merge now.

csadorf:
@CasperWA I'll squash and merge this PR after you approve.

aiidalab/app.py
CasperWA: So should we try to focus the information in this PR as well, or do you want to leave it for another go at it?

setup.json
```diff
@@ -14,10 +14,10 @@
     ],
     "extras_require": {
         "pre-commit": [
-            "pre-commit==1.17.0",
-            "yapf==0.28.0",
-            "prospector==1.1.7",
-      	    "pylint==2.3.1"
+		"pre-commit==1.17.0",
+		"yapf==0.29.0",
```
CasperWA, line 18: Newest version is 0.30, but not sure if this fits here?
```suggestion
		"yapf==0.30.0",
```

setup.json
```diff
@@ -14,10 +14,10 @@
     ],
     "extras_require": {
         "pre-commit": [
-            "pre-commit==1.17.0",
-            "yapf==0.28.0",
-            "prospector==1.1.7",
-      	    "pylint==2.3.1"
+		"pre-commit==1.17.0",
```
CasperWA, line 17: Newest version is 2.6.0. For the major version 1, the newest version is 1.21.0.
```suggestion
		"pre-commit==2.6.0",
```

setup.json
```diff
@@ -14,10 +14,10 @@
...
+		"yapf==0.29.0",
+		"prospector==1.3.0",
+		"pylint==2.5.2"
```
CasperWA, line 20:
```suggestion
		"pylint==2.5.3"
```

csadorf:
> Checked your updated dependencies, some are already out-of-date.
> Some you may not want to change according to my suggestions, but I think at least the `pylint` one could be updated :)

I bumped them specifically to meet the requirements of prospector version 1.3.0 to support dataclasses. I'd prefer to do a general bump of dependencies in a separate PR.

CasperWA:
> > Checked your updated dependencies, some are already out-of-date.
> > Some you may not want to change according to my suggestions, but I think at least the `pylint` one could be updated :)
> 
> I bumped them specifically to meet the requirements of prospector version 1.3.0 to support dataclasses. I'd prefer to do a general bump of dependencies in a separate PR.

So then there's only my question left on whether you want to address the `groups`/`categories` in this PR or another one?

aiidalab/app.py
csadorf: This data class is currently reflecting the information that is provided to the class. That information is partially redundant, but I'd prefer to address that issue in a separate change set.

csadorf:
> > > Checked your updated dependencies, some are already out-of-date.
> > > Some you may not want to change according to my suggestions, but I think at least the `pylint` one could be updated :)
> > 
> > 
> > I bumped them specifically to meet the requirements of prospector version 1.3.0 to support dataclasses. I'd prefer to do a general bump of dependencies in a separate PR.
> 
> So then there's only my question left on whether you want to address the `groups`/`categories` in this PR or another one?

I think addressing that might require some adjustments to the registry schema, which I'd prefer to address separately. See also my comment above.

aiidalab/app.py
yakutovicha: For the record: `groups` is an exact copy of `categories`. It was keept for the backwards compatibility. It is created [here](https://github.com/aiidalab/ansible-role-aiidalab-server/blob/9057f349f2a4fe5372628390d79a179079af60e5/files/merge_app_metadata.py#L109-L110)

PR #84, opened 2020-07-29
Make return value of AiidaLabApp.__repr__() valid Python expression.

csadorf:
@CasperWA I previously accepted your suggestion to add `<..>` delimiters to the `__repr__` return value of the `AiidaLabApp` class (as part of #81) however, the convention is not use these delimiters when `__repr__` actually returns a valid Python expressions, see also here:

 * https://docs.python.org/3.7/library/functions.html#repr
 * https://stackoverflow.com/a/1984177

This PR reverts that change.

### DIFF
--- a/aiidalab/app.py
+++ b/aiidalab/app.py
@@ -271,8 +271,8 @@ class AiidaLabApp(traitlets.HasTraits):
 
     def __repr__(self):
         app_data_argument = None if self._registry_data is None else asdict(self._registry_data)
-        return (f"<AiidaLabApp(name={self.name!r}, app_data={app_data_argument!r}, "
-                f"aiidalab_apps_path={os.path.dirname(self.path)!r})>")
+        return (f"AiidaLabApp(name={self.name!r}, app_data={app_data_argument!r}, "
+                f"aiidalab_apps_path={os.path.dirname(self.path)!r})")
 
     @traitlets.default('detached')
     def _default_detached(self):


CasperWA:
Well reading through your references, and also checking out the condition that it should be passable to [`eval()`](https://docs.python.org/3.7/library/functions.html#eval), I am not completely sure about this reversion.
Also (as a completely irrelevant but still "umbrella"-related things), we use angle brackets for the same thing in `aiida-core`, see, e.g., [here](https://github.com/aiidateam/aiida-core/blob/46af33022f49234307534c40089c0865103e4665/aiida/orm/nodes/node.py#L131) for the `Node` class, which all other `Node`s inherit from, and hence also this `__repr__` method.

As far as I understand it, the convention is to return something that can used in `eval()`, if it can not, then it should have angle brackets.

However, you can use `__str__` to return what you like :smiley: This should be the "standard" special method called for any actual purposes. Where `__repr__` is more of a "developers" insight into the object or class (as is also written on the stackoverflow page). And is indeed used as a fallback if no special `__str__` method can be found.

On the other hand. If you can have it return something that works with `eval()` that's cool with me :+1:

csadorf:
@CasperWA The `AiidaLabApp.__repr__()` return value is a valid expression. That was actually already intended when I made the change in #81, I just didn't pay enough attention when I accepted your suggestion.

You could use the following test code to convince yourself:
```python
from aiidalab.app import AiidaLabApp

app = AiidaLabApp(name='aiidalab-lsmo', app_data={'git_url': 'https://github.com/lsmo-epfl/aiidalab-epfl-lsmo.git', 'meta_url':'https://raw.githubusercontent.com/lsmo-epfl/aiidalab-epfl-lsmo/master/metadata.json', 'categories': ['classical','quantum'], 'groups': ['classical', 'quantum'], 'metainfo': {'description': 'Applications from the LSMO group', 'title': 'LSMO apps', 'version': '0.1.0', 'logo': 'miscellaneous/logos/LSMO.png', 'state': 'registered'}, 'gitinfo': {'HEAD': 'a86f4e32bf5b0bed10020322c6eb30d00205301d', 'refs/heads/che_609_course': '3ce85293d3649d326747544aa7570ca1651e6b38', 'refs/heads/develop':'a86f4e32bf5b0bed10020322c6eb30d00205301d', 'refs/heads/electronic_calcs': '77abb69ccb5642a7cdc9c9e29fcb9ba3e3842fab', 'refs/heads/master': 'cea77537e1f1f9b5172b559de6e7637cbb5e8257'}, 'hosted_on': 'github.com'}, aiidalab_apps_path='/home/aiida/apps')
app2 = eval(repr(app))
```

Overall, I am not quite sure how to interpret your comment. It seems like you are agreeing with my interpretation of the definition of `__repr__`, but then don't seem to come the same conclusion regarding this change... 😕 Re aiida-core, I think it's intentional that `Node.__repr__()` is not `eval`able, so very well in line with what I'm proposing here.

CasperWA:
Excellent :)

PR #88, opened 2020-07-30
Feature/cli

csadorf:


csadorf:
> I believe you should define a `cli` entry point. something like that:
> 
> ```json
>     "entry_points": {
>         "console_scripts": [
>             "aiidalab = aiidalab:cli"
>         ]
>     }
> ```

Yes, will add that.

> 
> Also, wouldn't it be better to name the file `cli.py` instead of `__main__.py`?

Using `__main__py` means that we can also run this with `python -m aiidalab` which is good for debugging.

PR #91, opened 2020-07-30
Move AppManagerWidget class into app_manager module.

csadorf:
To reduce the size of the app module.


aiidalab/app_manager.py
```diff
@@ -0,0 +1,292 @@
+# -*- coding: utf-8 -*-
+"""Module to manage AiiDA lab apps."""
```
csadorf, line 2: This must be updated.

PR #92, opened 2020-07-30
Run postBuild script after installing app environment.

csadorf:


csadorf:
After pondering over this for a little while, I'd like to withdraw this PR and put more thought into this, because we can't guarantee that changes made by these scripts are isolated to the app environment.

csadorf:
Replaced by #98 .

PR #97, opened 2020-08-10
Feature/app environments

csadorf:


csadorf:
I've tested this branch locally and it is incompatible with `aiidalab-home:develop` because the `install_app()` function was converted to a generator function. Before moving forward, I recommend that we implement a test that would catch this issue.

csadorf:
Test with: https://github.com/aiidalab/aiidalab-home/pull/39

yakutovicha:
@csadorf, I think we can safely close this PR, as we decided to go another route.

PR #98, opened 2020-08-10
Run postBuild script after installing app environment.

csadorf:


csadorf:
@yakutovicha I think it would make more sense to rename the `postBuild` script to `postInstall`, because while it would be nice to just replicate the myBinder setup, we are not doing exactly the same and to apply that makes not much sense IMO.

aiidalab/app.py
```diff
@@ -496,6 +509,17 @@ def install_environment(self):
                 self.environment.uninstall()  # rollback
                 raise RuntimeError(f"Failed to install app dependencies: {error.stderr.decode()}.")
 
+            else:
+                # ... then run the postBuild script (if present) ...
+                if Path(self.path).joinpath('postBuild').is_file():
```
yakutovicha, line 514: Why do you need to perform this check for the second time?

csadorf: This is a check whether we should execute the function. The assert within the function is just for safety in case that a developer or user calls the function otherwise. The `is_file()` check is extremely cheap in this context so duplication is a non-issue.

yakutovicha: Ok

PR #101, opened 2020-08-11
Cleanup/remove redundant dependencies

yakutovicha:


yakutovicha:
@csadorf I would consider moving all the widgets to the home app. That would allow us to remove `notebook` and `ipywidgets` from the dependency list.

csadorf:
> @csadorf I would consider moving all the widgets to the home app. That would allow us to remove `notebook` and `ipywidgets` from the dependency list.

I think that is generally a good idea. Do you want me to create a PR for that?

yakutovicha:
closing this PR as it is greatly outdated and it is not worth the effort to rebase it. I would rather open a new one.

PR #102, opened 2020-08-11
Fix error triggered by installing an app without dependencies.

csadorf:


csadorf:
This is already merged into the feature branch.

PR #103, opened 2020-08-11
Remove all widgets.

csadorf:
Moved to home app.


csadorf:
@yakutovicha The failing tests are expected, since we are testing with a home app version that does not support this. This is actually a good sing that tests failed.

csadorf:
> Looks good, to me, one question though: why did you remove `notebook` and `ipywidgets` from the dependency list? My understanding is that it can be done only after #97 is merged

You are correct. I'll revert that change.

csadorf:
@yakutovicha As expected, tests pass now.

PR #104, opened 2020-08-12
Automatically install app dependency into the app root directory.

csadorf:


csadorf:
For the record, the main issues with this approach are:

1. Using `--target=...` `pip` will ignore all previously installed packages, which poses a problem within this context. For example, `aiidalab-widgets-base` requires `aiida-core` which is actually installed in the system path, but using this approach would be installed again into the app dependency directory.
2. It is not clear that using this kind of "fake virtual environment", that additional commands would be properly executed, e.g., for the [installation of `ipywidgets`](https://ipywidgets.readthedocs.io/en/latest/user_install.html) we must execute `jupyter nbextension enable --py widgetsnbextension`, but since we do not have a dedicated kernel, this might lead to problems, especially when `ipywidgets` is installed multiple times for different apps at different locations.
3. While this approach uses a dedicated directory and symlinks in an effort to better contain the dependencies within the app root directory, the directory is still massively polluted with all of the dependency packages. App developers would probably need to adjust the`.gitignore` file to account for that, a dynamic update of that file would usually represent an app modification which must be avoided.

PR #106, opened 2020-08-13
Only require app name to construct AiidaLabApp instance.

csadorf:
That is the unique identifier of an app and should therefore be the only
information required to instantiate the object.

### DIFF
--- a/aiidalab/app.py
+++ b/aiidalab/app.py
@@ -11,7 +11,7 @@ from enum import Enum, auto
 from time import sleep
 from threading import Thread
 from subprocess import check_output, STDOUT
-from dataclasses import dataclass, field, asdict
+from dataclasses import dataclass, field
 
 from typing import List, Dict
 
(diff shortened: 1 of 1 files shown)


aiidalab/app.py
```diff
@@ -320,9 +316,10 @@ def current_revision(self):
 
             return None  # current revision not on the release line
 
-    def __init__(self, name, app_data, aiidalab_apps_path, watch=True):
+    def __init__(self, name, watch=True):
```
yakutovicha, line 319: I believe we should keep `app_data=None, aiidalab_apps_path=None` for some time. Otherwise, the non-updated home app will fail.

aiidalab/app.py
```diff
@@ -320,9 +316,10 @@ def current_revision(self):
...
         super().__init__()
 
+        app_data = load_app_registry()['apps'].get(name)
```
yakutovicha, line 322: Just to mention. Initially the idea was to load `app_data` only once to reduce the time needed. But, with caching enabled, everything seems to be fast here.

csadorf: Got it. I think the lesson here is to ensure to optimize at the right spot and also not prematurely. Code logic and clarity usually trumps micro optimizations.

aiidalab/app.py
csadorf: I think it is time that we properly address incompatibilites between the aiidalab system (that includes the `aiidalab` packages, but also all other packages installed into the system environment) and the dependents. Simply supporting a backwards compatible API "for some indefinite time" is simply not an option. I'll propose a plan for this.

PR #107, opened 2020-08-20
Testing/fix home app ci testing

csadorf:


csadorf:
@yakutovicha I'm honestly wondering whether the tests that we are running here make sense, because only a specific combination of this package and aiidalab-home are determined to be compatible. We can implement unit tests for the aiidalab package, but the kind platform tests don't make much sense.

yakutovicha:
> @yakutovicha I'm honestly wondering whether the tests that we are running here make sense, because only a specific combination of this package and aiidalab-home are determined to be compatible. We can implement unit tests for the aiidalab package, but the kind platform tests don't make much sense.

I agree, we don't need to test the platform here. We should only test the "core" features of the package.

csadorf:
> > @yakutovicha I'm honestly wondering whether the tests that we are running here make sense, because only a specific combination of this package and aiidalab-home are determined to be compatible. We can implement unit tests for the aiidalab package, but the kind platform tests don't make much sense.
> 
> I agree, we don't need to test the platform here. We should only test the "core" features of the package.

Ok, great. I'll do that and close this PR.

Issue #86, opened 2020-07-29
Install button is always active for the `aiidalab-qe` app.


yakutovicha:
After the app is installed the `Installed version` appears as unknown, while it is clearly the latest commint on the latest branch.

<img width="1141" alt="Screenshot 2020-07-29 at 16 47 42" src="https://user-images.githubusercontent.com/13118485/88815203-5b5afa80-d1bb-11ea-8c0d-07503046f756.png">

Pressing the `Install` button doesn't change anything.

csadorf:
I've removed the submodules on aiidalab-qe@master, so that will probably resolve the issue. I've tried to make dulwich properly parse the git submodule status, but that turns out to be a non-trivial undertaking that is probably not worth our time.

yakutovicha:
fixed in #114

Issue #87, opened 2020-07-29
Changing installed version results in "Unable to determine availability of updates" warning.


yakutovicha:
Even if I come back to the initially installed version, the warning remains.
<img width="1007" alt="Screenshot 2020-07-29 at 16 54 47" src="https://user-images.githubusercontent.com/13118485/88816029-416de780-d1bc-11ea-84bc-6146bf9d6a96.png">

yakutovicha:
closed in #114

PR #114, opened 2020-09-02
Revise installation and update implementation

csadorf:
This change set revises how the app installation, update determination, and app update flow is implemented and thus fixes issues #86 and #87. The change set also includes corresponding integration tests of these flows in various scenarios.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -12,6 +12,7 @@ jobs:
 
       - uses: actions/checkout@v2
         with:
+          fetch-depth: 0
           submodules: true
 
       - uses: actions/setup-python@v2
--- a/aiidalab/app.py
+++ b/aiidalab/app.py
@@ -407,11 +407,11 @@ class AiidaLabApp(traitlets.HasTraits):
                 check_output(['git', 'clone', url, self.path], cwd=os.path.dirname(self.path), stderr=STDOUT)
 
             # Switch to desired version
-            rev = self._release_line.resolve_revision(re.sub('git:', '', version))
+            rev = self._release_line.resolve_revision(re.sub('git:', '', version).encode())
             check_output(['git', 'checkout', '--force', rev], cwd=self.path, stderr=STDOUT)
 
             self.refresh()
-            return 'git:' + rev
+            return 'git:' + rev.decode()
 
     def update_app(self, _=None):
         """Perform app update."""
--- a/aiidalab/app.py
+++ b/aiidalab/app.py
@@ -10,7 +10,7 @@ from contextlib import contextmanager
 from enum import Enum, auto
 from time import sleep
 from threading import Thread
-from subprocess import check_output, STDOUT
+from subprocess import check_output, STDOUT, CalledProcessError
 from dataclasses import dataclass, field, asdict
 
 from typing import List, Dict
@@ -235,6 +235,11 @@ class AiidaLabApp(traitlets.HasTraits):
             This function returns None if the short-ref cannot be resolved
             to a full reference.
             """
+            # Check if short-ref is among the remote refs:
+            for ref in self._repo.refs.allkeys():
+                if re.match(r'refs\/remotes\/(.*)?\/' + short_ref, ref.decode()):
+                    return ref
+
             # Check if short-ref is a head (branch):
             if f'refs/heads/{short_ref}'.encode() in self._repo.refs.allkeys():
                 return f'refs/heads/{short_ref}'.encode()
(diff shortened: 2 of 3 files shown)


yakutovicha:
testing it.

yakutovicha:
If the app is not in the registry, does it suppose to have the "Latest Version": 
Screenshot 2020-09-03 at 10 45 55

csadorf:
> If the app is not in the registry, does it suppose to have the "Latest Version"?

No, I think it should say that it is not possible to determine whether it is latest version. I'll fix that.

yakutovicha (commented):
Another thing. I created a new branch starting from `master`. I didn't modify the new branch at all.

What I see is: the fact that it might be another branch is correctly recognized, while I cannot click on the `Install` button to bring things to the master branch.

Screenshot 2020-09-03 at 11 01 33

csadorf:
> Another thing. I created a new branch starting from `master`. I didn't modify the new branch at all.

I assume you meant to imply that you also checked out that other branch. The behavior is then as expected. Once a user leaves the "release line" by either checking out a different branch, or a commit which neither tagged on the branch or is the head commit, the UI capabilities are greatly reduced. The app is assumed to be in "developer" mode at that point, because it should be impossible to reach that state without manually manipulating the app state on the terminal.

yakutovicha:
> I assume you meant to imply that you also checked out that other branch.

right. 

>  The behavior is then as expected. Once a user leaves the "release line" by either checking out a different branch, or a commit which is neither tagged on the branch nor the branch's head commit, the UI capabilities are greatly reduced. 

I think users should be able to come back to the standard release line, no?

yakutovicha:
For some reason, I don't see the updates for the `aiidalab-widgets-base` package. The version `1.0.0b10` has been released a few days ago:

Screenshot 2020-09-03 at 12 08 33

yakutovicha:
I pulled the newest changes

For some reason, when I updated aiidalab-widgets-base to `1.0.0b9` version, I got back:

Screenshot 2020-09-03 at 12 13 07

CasperWA:
> > The behavior is then as expected. Once a user leaves the "release line" by either checking out a different branch, or a commit which is neither tagged on the branch nor the branch's head commit, the UI capabilities are greatly reduced.
> 
> I think users should be able to come back to the standard release line, no?

Well we could consider a "reset" button that only exists when the checkbox has been checked.
But otherwise I agree with @csadorf's conclusion here. If you as the user broke it, it's your responsibility to fix it. But a "reset to default/factory settings", might be an idea.

csadorf:
> For some reason, I don't see the updates for the `aiidalab-widgets-base` package. The version `1.0.0b10` has been released a few days ago:

The specific version (`1.0.0b10`) does not show up, because there is no way to figure out what versions are available without fetching the remote ref. That said, even if I reset my repository to not know about the upstream commit, the update button is enabled and it is indicated that there are updates available, and when I manually fetch, the version shows up, even if not installed. So I am not able to reproduce the issue on my end.

The *updates available* indication (via the label and the button) is enabled when one of the following two conditions is met:
1. Our local branch reference is out-of-date (as determined by comparing with the registry) and we need to fetch.
2. The installed version is not the latest on the specified release line (only possible for branches).

csadorf:
> > > The behavior is then as expected. Once a user leaves the "release line" by either checking out a different branch, or a commit which is neither tagged on the branch nor the branch's head commit, the UI capabilities are greatly reduced.
> > 
> > 
> > I think users should be able to come back to the standard release line, no?
> 
> Well we could consider a "reset" button that only exists when the checkbox has been checked.
> But otherwise I agree with @csadorf's conclusion here. If you as the user broke it, it's your responsibility to fix it. But a "reset to default/factory settings", might be an idea.

We can consider that, but that would not be in line with what we had agreed upon previously, and would anyways need to be implemented in the home app, so I suggest that we postpone that.

csadorf:
> I pulled the newest changes
> 
> For some reason, when I updated aiidalab-widgets-base to `1.0.0b9` version, I got back:
> 
> Screenshot 2020-09-03 at 12 13 07

Are you sure that your local copy of the app is pointing to the correct remote and is not otherwise modified? I am not able to reproduce the issue on my end.

yakutovicha:
> The specific version (`1.0.0b10`) does not show up, because there is no way to figure out what versions are available without fetching the remote ref.

The Git-related information is fetched automatically hourly, so you can find all the available tags there: https://aiidalab.materialscloud.org/appsdata/apps_meta.json

csadorf:
> > Edit: Btw, as a work-around until then, a simple reset should be to just uninstall and reinstall the app. :)
> 
> So we could, essentially, have a button that does this all-in-one?

Yes.

csadorf:
@yakutovicha This should be all set.

yakutovicha:
> @yakutovicha This should be all set.

I do a quick test and then will approuve

yakutovicha (approved):
Looks good, thanks @csadorf!

PR #115, opened 2020-09-03
Determine remote branch from release line instead of checked out branch.

csadorf:
Makes installation flow robust against detached HEAD states.

### DIFF
--- a/aiidalab/app.py
+++ b/aiidalab/app.py
@@ -451,26 +451,28 @@ class AiidaLabApp(traitlets.HasTraits):
     def _remote_update_available(self):
         """Check whether there are more commits at the origin (based on the registry)."""
         error_message_prefix = "Unable to determine whether remote update is available: "
-        try:
+
+        try:  # Obtain reference to git repository.
             repo = self._repo
         except NotGitRepository as error:
             raise AppRemoteUpdateError(f"{error_message_prefix}{error}")
-        try:
-            branch_ref = 'refs/heads/' + repo.branch().decode()
-            local_remote_ref = 'refs/remotes/origin/' + repo.branch().decode()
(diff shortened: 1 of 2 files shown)


yakutovicha (commented):
Alright, so two things I discovered:

1. The apps without a remote repo get green "Latest version" instead of the red "Unable to determine.."
2. If the app was in "Detached state", it always remains there, even after the update.

Technically, though, things work just fine.

csadorf:
I'll have a look at it. I'll need to investigate the failing test anyways. 

Just to confirm, would you prefer unregistered apps to always warn then?

yakutovicha:
> Just to confirm, would you prefer unregistered apps to always warn then?

The best would be to show the yellow text "Unregistered app". Otherwise, to be honest, both are fine with me: "Latest version" is kind of also true, because it is indeed the latest available version; the long scary message about "Inability to determine the updates" is fine as well. Do you have an opinion on this?

csadorf:
> > Just to confirm, would you prefer unregistered apps to always warn then?
> 
> The best would be to show the yellow text "Unregistered app". Otherwise, to be honest, both are fine with me: "Latest version" is kind of also true, because it is indeed the latest available version; the long scary message about "Inability to determine the updates" is fine as well. Do you have an opinion on this?

Since unregistered apps are considered development mode at the moment, I don't think it matters too much. Let's see how difficult it would be to display "Unregistered app", otherwise, I would suggest we leave it _as is_ and create an issue.

csadorf:
@yakutovicha I've fixed the issue pertaining to the failing test and cleaned up the function implementation a little bit. I am going to address "recovering" from a detached HEAD state in a separate PR.

yakutovicha (approved):

PR #118, opened 2020-09-08
Use the number sign '#' instead of '@' to indicate release line.

csadorf:
This is more in line with the URL specification (RFC 3986, 3.5) that
defines that the fragment section of a URL is not transmitted to the
server endpoint and evaluated on the client side to denote "some portion
or subset of the primary resource".

Using '#' instead of '@' also avoids potential confusion with the use of
the symbol in the context of `pip install`, which has highly related,
but not identical meaning.

### DIFF
--- a/aiidalab/app.py
+++ b/aiidalab/app.py
@@ -12,6 +12,7 @@ from time import sleep
 from threading import Thread
 from subprocess import check_output, STDOUT
 from dataclasses import dataclass, field, asdict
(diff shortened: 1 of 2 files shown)


csadorf:
There are no changes to the home app needed for this.

CasperWA (approved):
Looks good to me, thanks @csadorf !

I like the elegant ways of cutting down line numbers :+1:

yakutovicha (approved):

PR #129, opened 2020-10-09
Feature/aiidalab environment version

csadorf:
This changeset enables the specification of an app to environment compatibility map as part of the App metadata. The metadata is extended by the `aiidalab_environment_version` key which maps an app version to a compatible AiiDAlab environment version.

Here is a video demonstration as to how that looks in practice:
ezgif-6-f5d069d1089f

To test this feature in your own instance:
1. Switch into AiiDAlab developer mode as described [here](https://github.com/aiidalab/aiidalab/wiki#development-flow-for-simultaneous-back-end-and-front-end-changes-aiidalab-and-home-app) or use the [aiidalab-dev](https://github.com/aiidalab/aiidalab-dev).

1. Switch the home app to this feature branch: https://github.com/csadorf/aiidalab-home/tree/feature/app-compatibility-with-environment
1. Download a local version of the [`apps_meta.json`](https://aiidalab.materialscloud.org/appsdata/apps_meta.json) file into your home directory and add the `aiidalab_environment_version` field.
1. Apply the following patch to aiidalab to ensure that the local registry is used:
```diff --git a/aiidalab/config.py b/aiidalab/config.py
index b33407b..4d60a5a 100644
--- a/aiidalab/config.py
+++ b/aiidalab/config.py
@@ -4,6 +4,6 @@ from os import getenv
 AIIDALAB_HOME = getenv('AIIDALAB_HOME', '/project')
 AIIDALAB_APPS = getenv('AIIDALAB_APPS', '/project/apps')
 AIIDALAB_SCRIPTS = getenv('AIIDALAB_SCRIPTS', '/opt')
-AIIDALAB_REGISTRY = getenv('AIIDALAB_REGISTRY', 'https://aiidalab.materialscloud.org/appsdata/apps_meta.json')
+AIIDALAB_REGISTRY = getenv('AIIDALAB_REGISTRY', 'file:///home/aiida/apps_meta.json')
 AIIDALAB_DEFAULT_GIT_BRANCH = getenv('AIIDALAB_DEFAULT_GIT_BRANCH', 'master')
 AIIDALAB_ENVIRONMENT_VERSION = getenv('AIIDALAB_ENVIRONMENT_VERSION', '1.0.0')
```


csadorf:
Note that the CI tests are currently failing due to a bug in the latest aiidalab-docker-stack image.

csadorf:
Related to https://github.com/aiidalab/aiidalab-docker-stack/pull/128

aiidalab/app.py
```diff
@@ -527,17 +559,18 @@ def refresh_async(self):
     @property
     def metadata(self):
         """Return metadata dictionary. Give the priority to the local copy (better for the developers)."""
+        if self._registry_data is not None and self._registry_data.metainfo:
```
CasperWA, line 562: Just curiosity. Might there have been a case where an app is installed and this line would be true and this method therefore would return `dict(self._registry_data.metainfo)`, when it should have gone down to the `try/except` section?

CasperWA (approved):
Seems like all's good, cheers @csadorf !

The tests should succeed before merging, I guess, otherwise all good for me.

aiidalab/app.py
csadorf: I've rearranged that function to always prioritize the metadata which is published in the registry over the one that is in the app's repository. That is the only way to prevent that we lose our information concerning compatibility of newer versions once the app is installed. I think this is also much cleaner in terms of "single source of truth". 

The side-effect is that if you want to "override" an app's metadata you need to do so directly in a local registry and you can't do that by just changing the `metadata.json` in the app directory. That's slightly less convenient for development, but I don't really see a better way to do this. I consider this primarily as a "developer issue" so we can try to improve this, but I don't think we should prioritize it too much.

csadorf:
> Seems like all's good, cheers @csadorf !
> 
> The tests should succeed before merging, I guess, otherwise all good for me.

Thx. I'll try to fix the issue (https://github.com/aiidalab/aiidalab-docker-stack/issues/130) in a separate PR.

aiidalab/app.py
csadorf: One thing to keep in mind here is that once you start manipulating the `metadata.json` file in the app's directory you automatically leave the release line and the home app is going to start blocking the app management anyways. So even if we prioritized the local metadata that wouldn't really help much for development or testing of the aiidalab or the home app.

csadorf:
@CasperWA Sorry, your review was dismissed when I rebased.

yakutovicha:
I am also reviewing/testing. Just in case.

csadorf:
Looks like I forgot to push the latest commit on this branch and now tests are failing. I'm investigating.

csadorf:
@CasperWA @yakutovicha Tests are passing now.

aiidalab/app.py
```diff
@@ -503,13 +513,46 @@ def _installed_version(self):
...
+
+        def get_version_identifier(version):
+            "Get version identifier from version (e.g. git:refs/tags/v1.0.0 -> v1.0.0)."
+            if version.startswith('git:refs/tags/'):
+                return version[len('git:refs/tags/'):]
+            if version.startswith('git:refs/heads/'):
+                return version[len('git:refs/heads/'):]
+            if version.startswith('git:refs/remotes/'):  # remote branch
+                return re.sub(r'git:refs\/remotes\/(.+?)\/', '', version)
+            return version
```
yakutovicha, line 533: hehe, I was about to complain, but then I remembered the 2 rule ;)

yakutovicha (changes requested):
It looks like this PR introduces a problem related to the handling of the app versions.

The `aiidalab-widgets-base` app that has the latest commit to master tagged as `v1.0.0b12` is marked with "Unable to determine availability of updates".  Additionally, it shows on the appstore " The currently installed version of this app is not supported in this environment. "

To reproduce the problem just install the latest version of the `widgets-base` app.

yakutovicha:
@csadorf apart from the aforementioned issue, I have no more requests. Please fix it and proceed without an additional review from my side.

csadorf:
@yakutovicha The issue was caused by the fact that all versions of the base widgets app were prereleases. This is fixed now. I went ahead and merged this PR, but I'll keep working on this feature and we might need a few follow-up patches.

PR #130, opened 2020-10-12
Fix pytests

yakutovicha:
This PR fixes pytests by adding `pytest_tornasync` as a dev dependency.


csadorf:
Testing now.

csadorf (approved):
Approved as a temporary workaround. Let's make sure to create an issue documenting underlying cause and a more permanent long-term solution.

Side note: I'd prefer slightly more descriptive branch names, e.g., `fix/issue-130` or so that reference the issue number and/or reference the problem or the solution.

Edit: We could leave #131 open as an alternative to creating a new issue.

PR #133, opened 2020-10-16
Add specification of config in local config file.

csadorf:
This change sets makes it possible to configure aiidalab with a config file (`~/aiidalab.toml`). The primary use case is to be able to override some of the configuration values temporarily during development. However, the environment variables still take precedence by default unless the `develop = True` key is set. In that case aiidalab will actively warn the user with a little banner to ensure that we don't forget that the environment is modified:
Screen Shot 2020-10-16 at 4 29 55 PM

### DIFF
--- a/aiidalab/__init__.py
+++ b/aiidalab/__init__.py
@@ -1,3 +1,6 @@
 """AiiDAlab core tools."""
+from .config import _DEVELOP_MODE as DEVELOP_MODE
+
+__all__ = ['DEVELOP_MODE']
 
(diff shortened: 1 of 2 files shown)


csadorf:
@yakutovicha There appears to be some kind of issue related to the documentation build. Can you have a look at that?

csadorf:
> @yakutovicha There appears to be some kind of issue related to the documentation build. Can you have a look at that?

The issue is resolved.

yakutovicha (approved):
thanks @csadorf

PR #134, opened 2020-10-16
Revision of the app compatibility check

csadorf:
This change set enables compatibility determination before an app is installed and also fixes up some issues previously undetected related to #129 .


csadorf:
Converted draft as we will need to adjust the implementation of this feature to account for enabling deployment of AiiDAlab on the Quantum Mobile VM.

csadorf:
@yakutovicha I've implemented the discussed revisions.

Instead of using one monolithic `AIIDALAB_ENVIRONMENT_VERSION`, the app dependencies are now directly specified by package within the app metadata. Here is an example for the qe app:

```diff
@@ -206,7 +206,19 @@
         "version": "0.1.0",
         "authors": "Carl Simon Adorf, Aliaksandr Yakutovich",
         "logo": "miscellaneous/logos/QE.jpg",
-        "state": "development"
+        "state": "development",
+       "requires": {
+               ">=20.09.0": [
+                       "aiida-core~=1.4",
+                       "aiidalab-optimade~=1.1.2",
+                       "aiidalab-widgets-base==1.0.0b12",
+                       "aiida-quantumespresso~=3.2",
+                       "ase~=3.18",
+                       "fileupload==0.1.5",
+                       "nglview~=2.7.1",
+                       "ipywidgets~=7.5"
+               ]
+       }
       },
       "gitinfo": {
         "HEAD": "954989159d521f0881a79828c7addd9c876da7d4",
```
To test this I recommend to
1. Use the following `aiidalab.toml`:
```
develop = true
registry = "file:///home/aiida/apps_meta.json"
```
2. Download the `apps_meta.json` into the home directory.
3. Apply the patch above by storing it in a file `qe-app-reqs.patch` and then applying it with `patch apps_meta.json qe-app-reqs.patch`.
4. Test in combination with https://github.com/aiidalab/aiidalab-home/pull/47

yakutovicha (changes requested):
Thanks, @csadorf,

I noticed an issue here. When I set the requirements like so: 
```json
        "requires": {
                "<2020.10.0": [
                        "aiida-core~=1.4",
                        "aiidalab-optimade~=1.1.2",
                        "aiidalab-widgets-base==1.0.0b12",
                        "aiida-quantumespresso~=3.2",
                        "ase~=3.18",
                        "fileupload==0.1.5",
                        "nglview~=2.7.1",
                        "ipywidgets~=7.5"
                ],
                ">=2020.10.0": [
                        "aiida-core~=1.5",
                        "aiidalab-optimade~=1.1.2",
                        "aiidalab-widgets-base==1.0.0b12",
                        "aiida-quantumespresso~=3.2",
                        "ase~=3.18",
                        "fileupload==0.1.5",
                        "nglview~=2.7.1",
                        "ipywidgets~=7.5"
                ]
        }
```

I expect (unless I missed something) only `20.09.0` version to appear as the only version available to choose from. However, I got all of them:
Screenshot 2020-10-27 at 08 13 16

Also, I have a question: is it still possible to set the dependencies this way:
```json
        "requires": {
                "": [
                        "aiida-core~=1.4",
                        "ipywidgets~=7.5"
                ]
        }
```

Finally, concerning the name `requires` - I still think we should change it. Because soon we will add package installation, where we will need to redefine it again. Therefore, I would propose to rename it to something like `requires-installed` (or similar) clearly emphasizing that we are expecting those packages to be installed. Later, we can add something like `install-requires` to list the packages to be installed.

csadorf:
@yakutovicha In your requirements you use a major app version number of `2020`, whereas the actual app's major version number is `20`. Therefore, all app versions will fall into your first set. Everything is working as expected if you fix that.

It is still possible to use an empty specifier of `""` to denote *all versions*.

I think it is a good idea to start forward thinking about how we might want to distinguish between requirements that would need to be "present" and those that might need to be installed into a specific environment (again, I believe that we will need to strongly re-evaluate that approach and should not assume that this is the strategy we use moving forward (!)). 

However, I believe that there is a misconception here. The only reason to declare *any* requirements as part of the app's metadata as opposed to just maintaining them in a `requirements.txt` file or defining them as part of `setup.py` script is because we need to know and evaluate them *prior* to installing the app. By definition the app "cannot be installed" if any of those requirements is missing. Therefore, should we adopt an approach where we perform some post-download setup step where additional dependencies are installed (into a dedicated environment or not), those dependencies can and should be comfortably declared simply as part of the standard setup routine, i.e., as part of the app's `requirements.txt` or `setup.py` file. Therefore, I do not really see a reason for distinguishing these as part of the apps' metadata and would actually argue that declaring any dependencies that should be installed after download should not be declared there as to avoid over-specification.

I conclusion, while I don't anticipate that we ever need to declare dependencies that would be installed as part of the metadata, I support  that we rename the `requires` key to `install_requires` which makes clear that these are dependencies that need to be present **prior to installation**.

yakutovicha:
> @yakutovicha In your requirements you use a major app version number of `2020`, whereas the actual app's major version number is `20`. Therefore, all app versions will fall into your first set. Everything is working as expected if you fix that.

Right, sorry, my mistake.

> It is still possible to use an empty specifier of `""` to denote _all versions_.

Perfect

> I think it is a good idea to start forward thinking about how we might want to distinguish between requirements that would need to be "present" and those that might need to be installed into a specific environment (again, I believe that we will need to strongly re-evaluate that approach and should not assume that this is the strategy we use moving forward (!)).

sure

> However, I believe that there is a misconception here. The only reason to declare _any_ requirements as part of the app's metadata as opposed to just maintaining them in a `requirements.txt` file or defining them as part of `setup.py` script is because we need to know and evaluate them _prior_ to installing the app. By definition the app "cannot be installed" if any of those requirements is missing. Therefore, should we adopt an approach where we perform some post-download setup step where additional dependencies are installed (into a dedicated environment or not), those dependencies can and should be comfortably declared simply as part of the standard setup routine, i.e., as part of the app's `requirements.txt` or `setup.py` file. Therefore, I do not really see a reason for distinguishing these as part of the apps' metadata and would actually argue that declaring any dependencies that should be installed after download should not be declared there as to avoid over-specification.
> 
> I conclusion, while I don't anticipate that we ever need to declare dependencies that would be installed as part of the metadata, I support that we rename the `requires` key to `install_requires` which makes clear that these are dependencies that need to be present **prior to installation**.

Yes, thanks. Just one point here `install_requires` is used to specify the packages **to be installed** in the standard `setup.py` approach, therefore I was thinking to find another name to avoid confusions. That is why I was proposing `requires-installed` (or `installed_requires`).

yakutovicha (approved):
After a discussion with @csadorf we concluded that we keep the `requirements` specifications as is.

The reasons to do so are the following:

- Most probably we won't add a specification of packages to-be-installed in the `metadata.json` file. However, in case we do so, instead of doing like below:

```json
{
  "requires_installed": {"~=1.0": [...], "~=1.2": [...]},
  "install_requires": {">=1.0": [...], "~=1.2": [...]}
}
```

we better do something like this:
```json
{
  "requires": {"~=1.0": {"installed": [...], "install": [...]}}
}
```
As this approach appears to be more explicit and less error-prone (notice the `~=1.0` and `>=1.0` in the first example.


Finally, I would like to summarize the logic of how dependencies checks are performed:

-  if `requires` is not set, no checks applied. Technically, it defaults to `"": []` which means "all versions have no requirements".
- if `requires` **IS** set, all versions that are not within the specifications are not considered.
- if for the currently installed app version two requires-specifications are overlapping, then the dependency should fulfil at least one of them (logical OR). Implementing logical AND appears to be from one side complicated and from the other side, it is not super-clear how to define which version takes precedence in case of multiple overlaps.


Good job, @csadorf

PR #138, opened 2020-10-31
Move non-aiida related packages to the direct dependencies of aiidalab

yakutovicha:
The AiiDA plugins remain in the Pipfile, which is used by pipenv to
generate the complete environment to be put in requirements.txt.
The dependencies specified in the requirements.txt file are further set
as 'aiida' extra dependencies. The direct dependencies of `aiidalab` package
are moved to `install_requres`.


setup.py
```diff
@@ -16,6 +16,10 @@
     if requirements[0].startswith('-i'):
         requirements.pop(0)
 
+    print(requirements)
```
csadorf, line 19: That should probably removed, right?

setup.py
```diff
@@ -16,6 +16,10 @@
...
+
+    kwargs['extras_require']['aiida'] = requirements
```
csadorf, line 21: Wouldn't be many requirements duplicated in this way? Like `aiidalab-widgets-base` would appear here and in the basic aiidalab dependencies without extras.

csadorf (changes requested):
I think I'm a bit confused about the overall structure we are going for here. I understand that the aiidalab "core dependencies" are now defined as part of the `setup(install_requires=[]`) argument. All other requirements are added as an "aiida" extra dependency. However, since the aiidalab core dependencies are part of those, they are now defined twice. Is that acceptable?

Please confirm if this is correct:
1. If I want to setup my own AiiDAlab instance I will just `pip install aiidalab`.
2. If I want to install many likely required dependencies I use `pip install aiidalab[aiida]`.
3. For aiidalab-docker-stack we continue to use the `requirements.txt`.

I think my problem is the circular dependency definition we have here:
*core dependencies in setup.py -> defines Pipfile -> defines requirements.txt -> defines setup.py extras* .

setup.py
yakutovicha: yes, leftover from testing.

csadorf:
I think my proposed solution to the circular definition I complained about in my review is to define all extras actually within the `setup.py` (potentially by parsing a `setup.json`) and then placing them in the `Pipfile` *with* those extras, i.e.: `aiidalab = {editable = true, path = ".", extras=['aiida']}`

The `Pipfile` should then *only contain* that one line within the `packages`.

Finally, using this approach we use the `Pipfile` only to determine a complete consistent environment (work-around pip limitations), but not really to declare dependencies, so we could remove the `Pipfile` from the repository completely and instead determine the `requirements.txt` (lockfile) using [pip-tools](https://github.com/jazzband/pip-tools#requirements-from-setuppy) directly:
```
$ pip-compile > requirements.txt
```

yakutovicha:
> Please confirm if this is correct:
> 
> 1. If I want to setup my own AiiDAlab instance I will just `pip install aiidalab`.
> 2. If I want to install many likely required dependencies I use `pip install aiidalab[aiida]`.
> 3. For aiidalab-docker-stack we continue to use the `requirements.txt`.

My original intention was to use `pip install aiidalab[aiida]` to install the complete environment, but this doesn't work as I expected, unfortunately. The problem is, for example, if you say `bokeh>=1.4` in `install_requires` and in `requirements.txt` you put `bokeh==1.4.0`, then pip installs the version `2.2.1`, which basically means the `bokeh` dependency set in `extras_require` is kind of ignored (or just has a lower priority).

So I think the way you proposed to handle this makes more sense. So I'm going to implement this.

setup.py
```diff
@@ -4,22 +4,11 @@
 
 from setuptools import setup, find_packages
 
-if __name__ == '__main__':
+if __name__ in ["__main__", "builtins"]:
```
csadorf, line 7: I've seen this conditional in some other aiida-related packages and generally consider this bad practice. Why do we even need this?

yakutovicha: Otherwise `pip-compile` fails for me.

csadorf: My question is: why do we need "`if __name__ ... ` ?

yakutovicha: I never thought about this, honestly. Just copied what AiiDA had at the time (and still has btw).

csadorf: Unless there is a good reason I suggest we remove this.

yakutovicha: I quickly had a look here: https://stackoverflow.com/questions/59831397/run-setup-function-from-setuptools-only-if-name-main

Makes sense to me to remove it, as we are not putting to-be-imported functions in this file

yakutovicha: done

csadorf (approved):

csadorf (approved):

PR #143, opened 2020-11-10
Downgrade traitlets dependency to `~=4.0`

yakutovicha:
This is needed to work on Quantum Mobile where the standard Python
version is 3.6. The >5.0 version of traitlets does not support it.

### DIFF
--- a/setup.json
+++ b/setup.json
@@ -21,7 +21,7 @@
         "pip~=20.0",
         "requests-cache~=0.5.2",
         "toml~=0.10",
-        "traitlets~=5.0",
+        "traitlets~=4.0",
         "urllib3~=1.24",
         "watchdog~=0.10.2"
     ],


yakutovicha:
Mentioning @chrisjsewell as he reported the issue initially.

chrisjsewell:
Yep cheers, as mentioned in Ubuntu 18.06 of you apt install python3 you get python 3.6 as standard, so it will be a real issue to ensure python 3.7. 
In the AiiDAlab docker you essentially get round this by installing Conda, but that adds a whole extra layer of complexity for quantum mobile that we have thus far not needed

csadorf:
> Yep cheers, as mentioned in Ubuntu 18.06 if you apt install python3 you get python 3.6 as standard, so it will be a real issue to ensure python 3.7.
> In the AiiDAlab docker you essentially get round this by installing Conda, but that adds a whole extra layer of complexity for quantum mobile that we have thus far not needed

I think we certainly need to do some additional testing with traitlets 4.x before merging this. I will try to do that as part of my review.

_Further_: I think we can try to support both Python 3.6 and 3.7 for now, but we should try to move to homogenize the employed Python versions on QM and the docker stack **as soon as possible**.

Since there is no clear support schedule established yet, I suggest we try to converge onto Python 3.7 for now and then establish clear guidelines on what Python version(s) we support on what timeline moving forward.

csadorf (approved):
I did some basic testing, but in the end I figured it's actually better to just merge PRs like this and rely on our automated tests to catch if there is an issue (and expand them if they don't) rather than doing too much manual testing.

PR #162, opened 2021-03-22
🔧 MAINTAIN: Improve package

chrisjsewell:
before I even write my documentation, this package needs sorting out 😛


chrisjsewell:
I turned on https://docs.readthedocs.io/en/latest/pull-requests.html, but it does not seem to be triggering here yet 🤷

tox.ini
```diff
@@ -0,0 +1,31 @@
...
+
+# see also test/rmq/docker-compose.yml to start a rabbitmq server, required for the those tests
+
+[tox]
+envlist = py37
+
+[testenv]
+usedevelop = true
+
+[testenv:py{36,37,38,39}]
```
csadorf, line 15: I don't think we support all of these versions. I would stick to 37 and 38.

csadorf (approved):
Thx! 😄

tox.ini
chrisjsewell: removed 36 👍

chrisjsewell:
@csadorf in the commits just made, I moved to setup.cfg (using https://github.com/gvalkov/setuptools-py2cfg)
Note our very own @greschd was recently discussing some issues with obtaining the version in https://github.com/pypa/setuptools/issues/1724 😉 @greschd is this all fixed now?

chrisjsewell:
also added a LICENSE

csadorf (dismissed):

csadorf:
@chrisjsewell Let me know when this is ready to merge, please.

chrisjsewell:
yep all done cheers
I think the RTD CI should already be triggering, but we can figure that out after the merge

csadorf:
@chrisjsewell Can you fix the merge conflict and maybe enable push to fork by maintainers?

chrisjsewell:
> maybe enable push to fork by maintainers?

you mean this (it was already enabled)?
image

greschd:
> Note our very own @greschd was recently discussing some issues with obtaining the version in [pypa/setuptools#1724](https://github.com/pypa/setuptools/issues/1724) 😉 @greschd is this all fixed now?

All fixed is a bit of an overstatement -- there are two ways in which `setuptools` tries to get the version (don't remember in which order):

- Importing the module and accessing the attribute. This works as long as the `__version__` is defined before any missing dependency might hit. So if you put the `__version__` into a separate sub-module, make sure to import that one before anything else at the top-level.
- An AST-based finder, which parses the module in which `__version__` is defined. That calls `importlib.util.find_spec` on the module name though, which in turn imports the "parent" module. For example, if the `version is defined as `version = attr: aiidalab.some_sub_module.__version__`, it would load `aiidalab`. Of course this can again fail due to dependency problems.

But, TL;DR: You're all good here, because the `__version__` is defined in the top-level `__init__.py`.

csadorf:
> > maybe enable push to fork by maintainers?
> 
> you mean this (it was already enabled)?
> image

Sorry, my bad. I had trouble pushing, but something was wrong on my end.

csadorf (approved):

csadorf:
Thank you very much, @chrisjsewell !

docs/source/app_development/index.rst
```diff
@@ -15,8 +15,10 @@ Overview of the separate in-depth sections in this part of the documentation:
 
 To get an idea of how proper apps may be built and used, you can check out these examples:
 
-.. toctree::
-    :maxdepth: 1
+.. note::
+
+    AiiDAlab expects that app source code is managed with `git <https://git-scm.com/>`__.
+    To start tracking the app development with git, open the terminal once more, and execute:
```
yakutovicha, line 21: I am not sure what was the idea behind it, but this broke the docs.

csadorf: Looks like some weird merge conflict.

chrisjsewell: There was a merge conflict, whereby the pre-commit now removes trailing whitespace (that was present on line 19). Wouldn't have thought this should break the directive though 🤷

csadorf: Should be fixed in #166 .

PR #171, opened 2021-04-12
Added a more rigid test for the availability of the registry

espenfl:
The `requests` can fail, according to: https://docs.python-requests.org/en/latest/user/quickstart/#errors-and-exceptions.

Here we introduce a more rigid test, catching `requests.exceptions.RequestException` which covers all relevant exceptions. In addition, we also check that the exit code of `request` is 200 according to: https://docs.python-requests.org/en/master/user/quickstart/#json-response-content.

### DIFF
--- a/aiidalab/utils.py
+++ b/aiidalab/utils.py
@@ -51,9 +51,13 @@ def load_app_registry():
             return json.loads(file.read())
     else:
         try:
-            return requests.get(AIIDALAB_REGISTRY).json()
-        except ValueError:
+            registry = requests.get(AIIDALAB_REGISTRY)
+            # Now check also the exit code, only accept a 200 (OK, request succeeded)
+            assert registry.status_code == 200
+            return registry.json()
+        except requests.exceptions.RequestException as e:
             print("Registry server is unavailable! Can't check for the updates")
+            print(e)
             return dict(apps=dict(), catgories=dict())
 
 


aiidalab/utils.py
```diff
@@ -51,9 +51,13 @@ def load_app_registry():
             return json.loads(file.read())
     else:
         try:
-            return requests.get(AIIDALAB_REGISTRY).json()
-        except ValueError:
+            registry = requests.get(AIIDALAB_REGISTRY)
+            # Now check also the exit code, only accept a 200 (OK, request succeeded)
+            assert registry.status_code == 200
```
csadorf, line 56:
```suggestion
            registry.raise_for_status()
```

csadorf (approved):
Thank you for the contribution! I have one suggestion, but also fine to merge *as is*!

aiidalab/utils.py
csadorf: Actually, I'm not 100% sure that the way it's coded right now it would ever raise the intercepted exception, so this is actually more of a change request.

aiidalab/utils.py
```diff
@@ -51,9 +51,13 @@ def load_app_registry():
...
+            return registry.json()
+        except requests.exceptions.RequestException as e:
             print("Registry server is unavailable! Can't check for the updates")
+            print(e)
```
chrisjsewell, line 60: This exception string, will be very non-descriptive for the user.
If you want to include it, I would append it to the end of the current message.

chrisjsewell (changes requested):

csadorf:
Should be resolved by #175 .

PR #175, opened 2021-04-13
🐛 FIX: Unloadable home page when offline

chrisjsewell:
Because of this, if https://aiidalab.materialscloud.org/ is offline (as it is currently),
you can't even start aiidalab 😒.

I suggest  you also look through the rest of the code-base,
to check if there is anywhere else that assumes an internet connection.

### DIFF
--- a/aiidalab/utils.py
+++ b/aiidalab/utils.py
@@ -52,7 +52,7 @@ def load_app_registry():
     else:
         try:
             return requests.get(AIIDALAB_REGISTRY).json()
-        except ValueError:
+        except (ValueError, requests.ConnectionError):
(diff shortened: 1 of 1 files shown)


yakutovicha:
I think #171 is addressing exactly that.

chrisjsewell:
Well get it merged and released then lol, because currently it has completely broken aiidalab

chrisjsewell:
I assume related, in that when this is fixed, and the registry fails to load, the title of widgets does not show:

image

Surely these can be sourced from the widget itself?

csadorf (approved):

PR #178, opened 2021-05-18
Docs: add information on where to get the run.sh script.

yakutovicha:


mbercx:
Apparently this code block is not showing, probably because it's also indented? Maybe set up a RTD build for PR's so it's easy to check this :)

mbercx:
As a sidenote, I'd also add the docker pulls here maybe?

PR #182, opened 2021-06-02
Fix app installation issue

yakutovicha:


aiidalab/app.py
```diff
@@ -496,6 +496,10 @@ def refresh(self):
             with self.hold_trait_notifications():
                 self.available_versions = list(self._available_versions())
                 self.installed_version = self._installed_version()
+                if (
+                    not self.is_installed()
+                ):  # Do not check compatibility if the app is not installed.
+                    return
```
csadorf, line 502: Instead of returning the function here, we should move the compatibility check into the `else:` block (line 510/514).

csadorf (changes requested):

aiidalab/app.py
yakutovicha: The issue is that the `self._is_compatible(self.installed_version)` fails if `self.installe_version` is equal to "not installed"

yakutovicha: The other thing one can do is to add this check directly into the `self._is_compatible` function. What do you think?

csadorf: Yes, I think that was the original intention anyways. I think all we might need to do is remove the "raise" statement in https://github.com/aiidalab/aiidalab/blob/19f35375ffecec90ac969c7cebfef497b9d48f2f/aiidalab/app.py#L476 ? Looks like it was accidentally not removed.

yakutovicha: I can try that.

yakutovicha: yep, removing the `raise` seem to be enought.

csadorf (approved):

PR #184, opened 2021-06-04
Release/21.06.0

yakutovicha:


csadorf (commented):
Can you fix the package description and url in `setup.cfg` prior to release?

I don't think that "Meta package for the AiiDAlab python environment." is an accurate description and the url should also be updated.

csadorf (approved):

PR #187, opened 2021-07-12
Do not delete apps directly, but move them into a trash directory.

csadorf:
This is to work-around issues on some network file systems, that may
have lingering locks on files within the repository directory. The clean
up of the trash directory should be handled as a service.


yakutovicha (approved):
looks good to me.

PR #188, opened 2021-07-12
Allow including prereleases when finding matching releases.

csadorf:


yakutovicha (approved):
looks good, thanks @csadorf!

PR #198, opened 2021-08-03
Add a possibility to redirect dependency installation messages.

yakutovicha:


aiidalab/app.py
```diff
@@ -8,14 +8,14 @@
 import sys
 import tarfile
 import tempfile
+import subprocess
```
csadorf, line 11: Looks like this is not enforced yet, but imports should be alphabetically.

aiidalab/app.py
```diff
@@ -65,6 +65,7 @@ class _AiidaLabApp:
     logo: str = field(default_factory=str)
     categories: List[str] = field(default_factory=list)
     releases: dict = field(default_factory=dict)
+    stdout = sys.stdout
```
csadorf, line 68: Why are you adding this as a class attribute?

aiidalab/app.py
```diff
@@ -172,11 +173,19 @@ def _install_dependencies(self, python_bin):
...
+                bufsize=1,
+                stdout=subprocess.PIPE,
+                stderr=subprocess.STDOUT,
             )
+            while True:
+                output = process.stdout.readline()
+                if output == "" and process.poll() is not None:
+                    break
+                if output:
+                    self.stdout.write(output)
```
csadorf, line 188: Is this a recipe from SO or so? If yes, could you please add a comment with the link?

aiidalab/app.py
```diff
@@ -494,6 +503,9 @@ def _has_git_repo(self):
         except NotGitRepository:
             return False
 
+    def set_stdout(self, stdout):
+        self._app.stdout = stdout
+
```
csadorf, line 508: I don't think this should be a class attribute or property, but be passed directly to the function that uses it. Could you comment on the design choice, please?

csadorf (changes requested):

aiidalab/app.py
yakutovicha: done

aiidalab/app.py
yakutovicha: added comment.

aiidalab/app.py
yakutovicha: My motivation here was to limit the amount of code needed to achieve the goal of redirecting the output. The chain of function calls that would be needed to transfer the `stdout` arguments is quite long, IMO:

`AppManagerWidget._install_version` -> `AiidaLabApp.install_app()` -> `_AiidaLabApp.install()` -> `_AiidaLabApp._install_dependencies()` -> `_pip_install()`

Instead, I thought it would make sense to redirect all the kind-of-log messages to a single output stream once.

That was my motivation. @csadorf what are the advantages of passing stdout as a function argument?

aiidalab/app.py
yakutovicha: I answered this question in the comment below.

aiidalab/app.py
csadorf: Ok, so it should certainly not be a class attribute. 

Making it a class variable implies that it is part of the object's state, whereas in reality where you want to direct the output of a specific function calls is ephemeral and only relevant for the specific invocation of that function. You would also expect that the stream cannot be changed during the function execution, right? This design pattern also introduces potential race conditions, since you would need to lock the instance to ensure that the output stream is not changed while you are executing whatever function is requiring output.

In short, avoiding the chain of arguments is not a sufficient justification to introduce this pattern in my opinion.

So my recommendation would be to simply "bite the bullet" and pass the stream argument down the chain. Adding a simple `out=None` argument to the relevant functions is really not that complex. Should there be a second case where we need to redirect some stream, we could consider to refactor this for the `HasTrait` class, but at the moment I would vote against that.

aiidalab/app.py
csadorf: responded

aiidalab/app.py
yakutovicha: Ok, I agree with your reasoning.

csadorf (approved):
LGTM!

PR #204, opened 2021-09-09
update one-liner

ltalirz:
Update one-liner (slightly adapted from the paper to make the connection
to the apps).
Also, add a brief overview of technologies used.

### DIFF
--- a/docs/source/index.rst
+++ b/docs/source/index.rst
@@ -2,7 +2,14 @@
 Welcome to AiiDAlab's documentation!
 ####################################
 
-AiiDAlab is an environment for users with diverse expertise to access and run computational workflows embedded in AiiDAlab apps.
+AiiDAlab is a web platform that enables computational scientists to package scientific workflows and computational environments into "apps" and to share those apps with their collaborators and peers.
+
+AiiDAlab builds on
+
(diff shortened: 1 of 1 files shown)


docs/source/index.rst
```diff
@@ -2,7 +2,14 @@
...
 ####################################
 
-AiiDAlab is an environment for users with diverse expertise to access and run computational workflows embedded in AiiDAlab apps.
+AiiDAlab is a web platform that enables computational scientists to package scientific workflows and computational environments into "apps" and to share those apps with their collaborators and peers.
+
+AiiDAlab builds on
+
+   * `AiiDA <https://www.aiida.net/>`_ as the computational workflow engine
+   * the `Jupyter <https://jupyter.org/>`_ environment (notebooks, widgets, ...) for writing and sharing apps
+   * `GitHub <http://github.com/>`_ for collaboration on app development
```
csadorf, line 11: This one is a bit of a half-truth. You don't have to share apps via GitHub, any public repository (does not even have to be git) is sufficient. I would recommend to remove that.

csadorf (changes requested):

docs/source/index.rst
ltalirz: fair, I'll remove it

PR #205, opened 2021-09-10
Revise section on app development

ltalirz:
The section on app development is revised.

 * Remove explanations of individual sections in the index.rst - just use self-explanatory titles
 * Remove "build your first app" - after templating, the next step should be to add widgets
 * move "publish app" to last point
 * rename QE example to "examples" - I think we should simply link to examples that people may want to check out


ltalirz:
The old structure of the "App development" page leads to a very confusing toctree

image

I propose to make the page titles more self-explanatory so that the "explanation of what each page does" becomes obsolete

ltalirz:
@csadorf there is more to do but I think this is ready for review

docs/source/app_development/create.rst
```diff
@@ -135,4 +135,6 @@ Rename the file to ``main`` and then add the following line into the first cell:
 Save the notebook and close it.
 Now go back to AiiDAlab start page and click on the ``My App`` link again - it should bring you to a page that says "Welcome to my app!"
 
-**Next:** :ref:`develop-apps:first-app`
+ In case you encounter any issues, compare your app to the source code of the `hello world app <https://github.com/aiidalab/aiidalab-hello-world>`_ that was templated by the cookiecutter approach.
```
csadorf, line 138: Is the indentation intentional?

docs/source/app_development/examples.rst
```diff
@@ -1,8 +1,8 @@
...
+.. _develop-apps:examples:
 
-****************************************************
-Build an app - AiiDA Quantum ESPRESSO bands workflow
-****************************************************
+************
+Example apps
+************
 
 Here we will have an example of how to build an AiiDAlab app that utilizes `AiiDA <https://www.aiida.net>`__ with the `AiiDA-Quantum ESPRESSO plugin <https://aiida-quantumespresso.readthedocs.io/>`__ to calculate the band structure for a compound.
```
csadorf, line 7: I think writing a QE app (even a simple one) as an example is total overkill. I would suggest that we change this to describe something much simpler.

docs/source/app_development/widgets.rst
```diff
@@ -1,7 +1,7 @@
 .. _develop-apps:widgets:
 
 ########################
-Widgets for AiiDAlab app
+Using AiiDAlab widgets
```
csadorf, line 4: Please make sure to update the length over the over- and underline.

csadorf (changes requested):

docs/source/app_development/create.rst
ltalirz: no it wasn't - thanks for checking

docs/source/app_development/examples.rst
ltalirz: What if we just start with a set of links to example apps?
This would little effort and probably very helpful?

That was the thought behind the change of the title - I'll change the text as well

ltalirz:
@csadorf  I've added the example apps for now; feel free to edit/update the list if you'd like to see other choices

PR #206, opened 2021-09-11
update widgets docs

ltalirz:
* reorder docs according to priority of information (start with aiidalab base widgets, then provide links to more examples, then add some background info for those who are interested)
* revise language
* remove js/python code illustrations for custom widgets - this bloats
  this page, would need to be maintained, and should be left to
  the ipywidgets docs

### DIFF
--- a/docs/source/app_development/widgets.rst
+++ b/docs/source/app_development/widgets.rst
@@ -11,7 +11,7 @@ We have created a collection of reusable widgets that are already integrated wit
 AiiDAlab widgets
 ****************
 
-AiiDAlab apps typically involve some of following steps:
+AiiDAlab apps typically involve some of the following steps:
 
  * prepare the input for a calculation (e.g. an atomic structure)
  * select computational resources and submit a calculation to AiiDA
@@ -49,7 +49,7 @@ Widgets and traitlets
 Widgets are eventful Python objects that display GUI components, such as a sliders, buttons, progress bars, drowdowns, etc.
 `ipywidgets`_ is a Python package that provides interactive widgets for the use in Jupyter notebooks.
 
-An Ipywidget has a parameter, whose value can be ovserved and accessed from Python.
+Widgets can have one or more attributes, whose value can be _ovserved_ and accessed from Python such that we can react to changes to their values.
 This is implemented using `traitlets <https://traitlets.readthedocs.io/>`_.
 
 For example, try the following code in a Jupyter notebook:
@@ -79,13 +79,13 @@ For example, try the following code in a Jupyter notebook:
 Creating your own widgets
 =========================
 
-Each ipywidget consists of a Python component that defines how to interact with the widget from Python,
+Each widget consists of a Python component that defines how to interact with the widget from Python,
 and a Javascript (or TypeScript) component that is responsible for the graphical representation of the widget and communicates updates back to the Jupyter kernel.
 
-If your goal is to combine existing widgets into new, reusable components, this can be done in Python without touching the Javascript component.
-See e.g. the `implementation of the AiiDAlab widgets <https://github.com/aiidalab/aiidalab-widgets-base>`_, most of which are of this type.
+If the goal is to combine existing widgets into new, reusable components, this can be done in Python without touching the Javascript component.
+See for example the `implementation of the AiiDAlab widgets <https://github.com/aiidalab/aiidalab-widgets-base>`_, most of which are of this type.
 
-If you would like to modify the appearance of an existing widget or create an entirely new visualization, you will need to write Javascript/TypeScript.
+To modify the appearance of an existing widget or to create an entirely new visualization, one needs to write Javascript/TypeScript.
 See the detailed `tutorial <https://ipywidgets.readthedocs.io/en/stable/examples/Widget%20Custom.html>`_ on how to develop a custom widget or have a look at some of the examples from :ref:`develop-apps:widgets:more-widgets`.
 
 .. _ipywidgets: https://ipywidgets.readthedocs.io
--- a/docs/source/app_development/widgets.rst
+++ b/docs/source/app_development/widgets.rst
@@ -1,106 +1,91 @@
 .. _develop-apps:widgets:
 
-######################
-Using AiiDAlab widgets
-######################
+####################
+Use AiiDAlab widgets
+####################
 
-AiiDAlab largely dependents on Jupyter widgets for the graphical user interface (GUI).
-More specifically, the `Ipywidgets <https://ipywidgets.readthedocs.io/>`_ package is the primary tool for creating AiiDAlab apps.
-We have also created the `aiidalab-widgets-base <https://github.com/aiidalab/aiidalab-widgets-base>`_ that is based on `Ipywidgets` and provides tools to interact with AiiDA.
-
-**********
-Ipywidgets
-**********
-
-`Ipywidgets <https://ipywidgets.readthedocs.io/>`_ is a python package that provides interactive widgets to be used in Jupyter notebooks.
-Widgets are eventful python objects, which shows GUI components such as a slider, button, etc.
-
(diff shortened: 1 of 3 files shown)


docs/source/app_development/widgets.rst
```diff
@@ -1,106 +1,91 @@
...
-**********
-Ipywidgets
-**********
+****************
+AiiDAlab widgets
+****************
 
-`Ipywidgets <https://ipywidgets.readthedocs.io/>`_ is a python package that provides interactive widgets to be used in Jupyter notebooks.
-Widgets are eventful python objects, which shows GUI components such as a slider, button, etc.
+AiiDAlab apps typically involve some of following steps:
```
csadorf, line 14:
```suggestion
AiiDAlab apps typically involve some of the following steps:
```

docs/source/app_development/widgets.rst
```diff
@@ -1,106 +1,91 @@
...
 
-Ipywidgets package provides base widgets like sliders, progress bar, buttons, checkbox, dropdown, radio buttons, text, text area, label, etc.
-The widgets have properties and value, which can be connected to the python codes and updated dynamically.
-For a simple example, one can use a slider to control the value of a variable.
+ * prepare the input for a calculation (e.g. an atomic structure)
+ * select computational resources and submit a calculation to AiiDA
+ * monitor a running calculation
+ * find and analyze the results of a calculation
```
csadorf, line 19: These are sentences.

docs/source/app_development/widgets.rst
```diff
@@ -1,106 +1,91 @@
...
-Ipywidgets package also supports developers in creating their own widgets.
-There is a detailed `tutorial <https://ipywidgets.readthedocs.io/en/stable/examples/Widget%20Custom.html>`_ to show how to develop a custom widget.
-In the ipywidgets framework is based on the Comm framework.
-In the Comm framework, the change events of the traits leads to send a json massage to the Jupyter kernel.
-Then, the frontend will update accordingly.
+Widgets and traitlets
+======================
 
-Develop the custom widgets mainly involve two parts:
+Widgets are eventful Python objects that display GUI components, such as a sliders, buttons, progress bars, drowdowns, etc.
```
csadorf, line 49: "Eventful", or "stateful"? I actually never heard the former term. :/

docs/source/app_development/widgets.rst
```diff
@@ -1,106 +1,91 @@
...
+`ipywidgets`_ is a Python package that provides interactive widgets for the use in Jupyter notebooks.
 
-* **Python**: in the Python part, one needs to define the APIs of the widgets.
-  For instance, the color of the widget's background.
-  The traits types should be initialized as:
+An Ipywidget has a parameter, whose value can be ovserved and accessed from Python.
```
csadorf, line 52:
```suggestion
Widgets can have one or more attributes, whose value can be _ovserved_ and accessed from Python such that we can react to changes to their values.
```

docs/source/app_development/widgets.rst
```diff
@@ -1,106 +1,91 @@
...
-.. code-block:: javascript
+        This function is called when the slider value is changed by the user.
+        """
+        print(change["old"])
+        print(change["new"])
+        if change["new"] > 373.0:
+            print("Boiling now!")
 
-    this.model.on('change:color', this._color_change);
+    slider.observe(slider_change, names="value")
```
csadorf, line 77: I am usually not a fan to maintain documentation for other packages here, since APIs and best practices may change. If we do so, I'd prefer with as little code as possible to make it a true "minimal example", I believe this could be achieved by removing the print function calls in lines 72 and 73.

docs/source/app_development/widgets.rst
```diff
@@ -1,106 +1,91 @@
...
 
-Here is a list for some useful custom widgets:
+Creating your own widgets
+=========================
 
-* `widget-periodictable <https://github.com/osscar-org/widget-periodictable>`_ : A interactive periodic table.
-* `widget-bandsplot <https://github.com/osscar-org/widget-bandsplot>`_ : A widget to plot bandstructure and density of states.
-* `widget-jsmol <https://github.com/osscar-org/widget-jsmol>`_ : A widget to use the molecular visualizer Jmol inside the Jupyter.
+Each ipywidget consists of a Python component that defines how to interact with the widget from Python,
```
csadorf, line 82:
```suggestion
Each widget consists of a Python component that defines how to interact with the widget from Python,
```
I don't think it makes sense to keep calling them "ipywidgets". That's the name of the packge, but the component is simply a widget. See also the language here: https://ipywidgets.readthedocs.io/en/stable/examples/Widget%20Basics.html

docs/source/app_development/widgets.rst
```diff
@@ -1,106 +1,91 @@
...
+and a Javascript (or TypeScript) component that is responsible for the graphical representation of the widget and communicates updates back to the Jupyter kernel.
 
-*********************
-AiiDAlab base widgets
-*********************
+If your goal is to combine existing widgets into new, reusable components, this can be done in Python without touching the Javascript component.
```
csadorf, line 85:
```suggestion
If the goal is to combine existing widgets into new, reusable components, this can be done in Python without touching the Javascript component.
```
Please do not directly address the reader. The only exception would be ("see ... ").

docs/source/app_development/widgets.rst
```diff
@@ -1,106 +1,91 @@
...
+See e.g. the `implementation of the AiiDAlab widgets <https://github.com/aiidalab/aiidalab-widgets-base>`_, most of which are of this type.
```
csadorf, line 86:
```suggestion
See for example the `implementation of the AiiDAlab widgets <https://github.com/aiidalab/aiidalab-widgets-base>`_, most of which are of this type.
```

docs/source/app_development/widgets.rst
```diff
@@ -1,106 +1,91 @@
...
 
-Molecular and material simulations always need database parser, molecular visulizer and editor, image render tools etc.
-Here, we privde the `aiidalab-widgets-base <https://github.com/aiidalab/aiidalab-widgets-base>`_ , which is a collection of these tools.
-Developers can easily reuse it to develop AiiDAlab apps.
-Read more information at `https://aiidalab-widgets-base.readthedocs.io <https://aiidalab-widgets-base.readthedocs.io/>`_ .
+If you would like to modify the appearance of an existing widget or create an entirely new visualization, you will need to write Javascript/TypeScript.
```
csadorf, line 88:
```suggestion
To modify the appearance of an existing widget or to create an entirely new visualization, one needs to write Javascript/TypeScript.
```

csadorf (changes requested):

docs/source/app_development/widgets.rst
ltalirz: This adjective was already there before; I'll just remove it

docs/source/app_development/widgets.rst
ltalirz: I generally agree, which is why I removed the more involved python+JS example and only left this one in.

Will also remove lines 72+73

docs/source/app_development/widgets.rst
csadorf: I figured, but wanted to raise that point regardless.

docs/source/app_development/widgets.rst
```diff
@@ -49,7 +49,7 @@ Widgets and traitlets
 Widgets are eventful Python objects that display GUI components, such as a sliders, buttons, progress bars, drowdowns, etc.
 `ipywidgets`_ is a Python package that provides interactive widgets for the use in Jupyter notebooks.
 
-An Ipywidget has a parameter, whose value can be ovserved and accessed from Python.
+Widgets can have one or more attributes, whose value can be _ovserved_ and accessed from Python such that we can react to changes to their values.
```
csadorf, line 52:
```suggestion
Widgets can have one or more attributes, whose value can be _observed_ and accessed from Python such that we can react to changes to their values.
```
typo

csadorf (approved):
One typo to fix, otherwise good to merge.

PR #222, opened 2021-09-22
Distinguish between non-registered and registry cannot be reached.

### LINKED ISSUE #221: Improve indication about whether an app is not registered or the registry server cannot be reached
Currently the app widget has a `updates_available` trait that can be in three states:
 - None: unable to determine, i.e., the registry server cannot be reached
 - True: update is available
 - False: update is not available,  i.e., the app is up-to-date

The proposed solution would be to add a 4th state _"registry server is reachable, but app is not registered"_ to better handle this as part of the app management in the home app.

Related to: #216 .

csadorf:
- The state where the registry cannot be reached and where an app is
not registered are clearly distinguished. It should be noted that when
the registry cannot be reached, it is not possible to determine whether
an app is hypothetically registered or not.
 - Replace `updates_available` trait with `remote_update_status`.
   To support more than three states.
 - Refactor version determination.

Fixes #221.

### DIFF
--- a/aiidalab/__main__.py
+++ b/aiidalab/__main__.py
@@ -108,7 +108,10 @@ def search(app_query, prereleases):
         try:
             app_requirements = [Requirement(app_query)]
         except InvalidRequirement:  # interpreted as general search query
-            registry = load_app_registry_index()
+            try:
+                registry = load_app_registry_index()
+            except RuntimeError as error:
+                raise click.ClickException(error)
             app_requirements = [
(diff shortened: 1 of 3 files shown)


csadorf:
Test with https://github.com/aiidalab/aiidalab-home/pull/71 .

yakutovicha (approved):
Thanks, @csadorf.

I think with this PR the logic looks considerably clearer.

I tested things locally and didn't encounter any issues.

PR #225, opened 2021-09-23
Add function to build registry

csadorf:
- Adds the `aiidalab registry` command group with registry related functions.
- Enables building of the app registry directly with `aiidalab registry build`.
- The `parse_app_repo` command is moved into the `aiidalab registry` command group.
- The API v0 is dropped and API version 1 is revised to version 1.1 in a backwards-compatible manner.
- API v1.1 adds per-release metadata to the app entries.

Note: This patch introduces a new schema version (however, that is largely due to revisions of the apps index), the API is updated to version 1.1 in a backwards-compatible manner.


csadorf:
See https://github.com/aiidalab/aiidalab-registry/tree/use-aiidalab-package for an example on how this can be used to host a registry.

csadorf:
Test in combination with: https://github.com/aiidalab/aiidalab-app-cutter/tree/setup-cfg

yakutovicha (commented):
Thanks, @csadorf. I tried to build and serve the registry, and it worked as expected.

However, I have a small confusion with handling schema versions. If I understand correctly, one cannot choose which version should be built. Is that correct? It looks that the mocking works for the `v2` only. Would it be possible to add an option for specifying which schema type to employ?

Another idea that I had is: can we replace the links in the `schema.json` files with the relative paths to the files? I think it would be nice because then we can avoid mocking. Also, implementing the schema selection that I mentioned earlier would become easier.

csadorf:
> However, I have a small confusion with handling schema versions. If I understand correctly, one cannot choose which version should be built. Is that correct? It looks that the mocking works for the `v2` only. Would it be possible to add an option for specifying which schema type to employ?

This registry expects the app index to adhere to schema version 2 and version 2 only. The JSON files that are built and served as part of the API also adhere to version 2. The changes that are thus introduced to the API are reflected by the fact that the API version number was increased from 1.0 to 1.1. The change in itself should be backwards-compatible since their are additive, with one exception, the deprecated `apps_meta.json` endpoint was removed.

I am therefore a bit at loss what you mean by specifying which schema type to employ. Do you mean an option to serve either API version 1.0 or 1.1?
 
> Another idea that I had is: can we replace the links in the `schema.json` files with the relative paths to the files? I think it would be nice because then we can avoid mocking. Also, implementing the schema selection that I mentioned earlier would become easier.

I was under the impression that `$ref` links cannot be relative paths, however I might be mistaken and I will look into this. Besides that though, is there an issue with mocking during the development and testing phase?

csadorf:
@yakutovicha According to the [JSON-schema reference documentation](https://json-schema.org/understanding-json-schema/structuring.html#id) it is not recommended to use relative references for the `$id` of the _base schema_. However, the same does not hold true for sub-schemas, meaning schemas that are merely referenced by other base schemas. In our case, `app.schema.json`, `apps.schema.json`, and `apps_index.schema.json` are base schemas, all others are sub-schemas. I have therefore removed the `$id` value from all sub-schemas in d39691ee1c900b05cbd9e060e5011384778adea6 which allows them to be referenced via a relative path in the base schemas.

That said, the topic of schema `$ids` and how [exactly references are resolved is relatively complex](https://json-schema.org/understanding-json-schema/structuring.html#ref). It does depend on the schema's base URL which for relative URIs would depend on how a particular schema was retrieved. The practice of preloading schemas into some local schema store like we do in our mocking code path is apparently not atypical.

I would recommend to stick to the current approach with absolute URIs for the base schema ids and otherwise use relative paths where possible. The practice of mocking those absolute URIs during testing and development seems robust enough to me.

yakutovicha:
> This registry expects the app index to adhere to schema version 2 and version 2 only. The JSON files that are built and served as part of the API also adhere to version 2. The changes that are thus introduced to the API are reflected by the fact that the API version number was increased from 1.0 to 1.1. The change in itself should be backwards-compatible since their are additive, with one exception, the deprecated `apps_meta.json` endpoint was removed.

Alright, so following this I can guess that the schema version 1 is not used at all, is that correct? Because if this is correct, then schema v1 is not needed and can be removed. Herein, if it can be removed, it probably doesn't make sense to version the schemas and they can be tightly bound to the API version?

If my logical flow above is correct, the only thing we ultimately care about is the API version number. Following this, we could probably collect this information in one place (probably a folder) and reference it while building the registry.

csadorf:
> > This registry expects the app index to adhere to schema version 2 and version 2 only. The JSON files that are built and served as part of the API also adhere to version 2. The changes that are thus introduced to the API are reflected by the fact that the API version number was increased from 1.0 to 1.1. The change in itself should be backwards-compatible since their are additive, with one exception, the deprecated `apps_meta.json` endpoint was removed.
> 
> Alright, so following this I can guess that the schema version 1 is not used at all, is that correct? Because if this is correct, then schema v1 is not needed and can be removed. Herein, if it can be removed, it probably doesn't make sense to version the schemas and they can be tightly bound to the API version?
> 

Yes, I don't think it makes sense to bind it to the API version, I would just bind it to the package version.

> If my logical flow above is correct, the only thing we ultimately care about is the API version number. Following this, we could probably collect this information in one place (probably a folder) and reference it while building the registry.

Not sure what you mean by that.

yakutovicha (approved):

PR #233, opened 2021-10-01
Allow control of verbosity for CLI and set default level to warning.

csadorf:


aiidalab/__main__.py
```diff
@@ -72,8 +69,9 @@ def _list_apps(apps_path):
 
 @click.group()
 @click.version_option(version=__version__, prog_name="AiiDAlab")
-def cli():
-    pass
+@click.option("-v", "--verbose", count=True)
+def cli(verbose):
+    logging.basicConfig(level=logging.WARNING - 10 * verbose)
```
yakutovicha, line 74: thanks, @csadorf. Just one question: what happens if the level value is negative?

yakutovicha (approved):
One question, otherwise good to go.

aiidalab/__main__.py
csadorf: Just ran the following minimal demo:
```python
import logging

logging.basicConfig(level=-1)
logger = logging.getLogger('test')

logger.info("hello")
```
and it's fine, however, I suggest that we simply safeguard against potential undefined behavior with:
```suggestion
    logging.basicConfig(level=max(0, logging.WARNING - 10 * verbose))
```

yakutovicha:
@csadorf do you want to merge this?

PR #234, opened 2021-10-01
docs: update admin section

ltalirz:
* Readers should be pointed to "quick" solutions before reading
   longer ones
 * Do not encourage admins to create their own registries


csadorf (approved):
Great, thanks a lot!

PR #235, opened 2021-10-04
DOC: note for how to develop when UID of host and container different

unkcpz:
A note is added to app create Step 5, which describe how to add code to
make the app functional. When running aiidalab locally, the UID of
aiidalab's bind mounted home folder may different from the UID of host
user. This lead to editing files inside folder permission denied because
of the ownership issue.
Vscode has a feature "Attach to a running container" to make
development inside the container possible.


docs/source/app_development/create.rst
```diff
@@ -204,4 +204,11 @@ Typically one would prefer to use the cookie-cutter to create a new app, but her
 
    In case you encounter any issues, compare your app to the source code of the `hello world app <https://github.com/aiidalab/aiidalab-hello-world>`_ that was templated by the cookiecutter approach.
 
+  .. note::
```
csadorf, line 207: I would suggest to show this as a tip.
```suggestion
  .. tip::
```

docs/source/app_development/create.rst
```diff
@@ -204,4 +204,11 @@ Typically one would prefer to use the cookie-cutter to create a new app, but her
...
+
+       The UID and GID of folder bind to container may different from the user's UID and GID in host machine,
+       Which will lead to edit files in bind mount aiidalab folder permission denied.  
+       If you are using vscode as IDE, there is a `Attach to a running container <https://code.visualstudio.com/docs/remote/attach-container>` 
+       feature that allows user to edit files inside the container without mess the onwership of files.
```
csadorf, line 212: Please use convention *one sentence per line*.

docs/source/app_development/create.rst
```diff
@@ -204,4 +204,11 @@ Typically one would prefer to use the cookie-cutter to create a new app, but her
 
    In case you encounter any issues, compare your app to the source code of the `hello world app <https://github.com/aiidalab/aiidalab-hello-world>`_ that was templated by the cookiecutter approach.
 
+  .. note::
+
+       The UID and GID of folder bind to container may different from the user's UID and GID in host machine,
+       Which will lead to edit files in bind mount aiidalab folder permission denied.  
+       If you are using vscode as IDE, there is a `Attach to a running container <https://code.visualstudio.com/docs/remote/attach-container>`
```
csadorf, line 211: Please fix the hyperlink, this is [restructured text](https://www.sphinx-doc.org/en/master/usage/restructuredtext/basics.html), not markdown.

csadorf (changes requested):
Thanks a lot for the PR! 👋 

1. Please make sure to run the pre-commit checks, since they are currently failing.
2. I checked https://aiidalab--235.org.readthedocs.build/en/235/app_development/create.html and it looks like the document is not properly formatted since the layout is all messed up. Can you try to fix that, please? 
3. I think it would be nice to actually provide a little bit more detail. Or is the process extremely straight-forward? 
4. I would place this tip much higher on the page since most people might not even read all the way to the bottom since "creating the app from scratch" is generally discouraged (developers should use the cookiecutter instead). Right above "Create app with cookiecutter" would probably be the right spot.

Issue #236, opened 2021-10-04
When updating the package "Include prereleases" checkbox value is not taken into account.


yakutovicha:
Currently, running an update always leads to the installation of the latest version, even if it is a pre-release.

PR #237, opened 2021-10-06
Fix: Correctly handle updates for apps with prereleases

csadorf:
Unless a prerelease is already installed or explicit user opt-in.

Note: this fixes a regression that was introduced with #222.


csadorf:
> Thanks, @csadorf. I tested and it looks good.
> 
> Can you maybe also look into #236? I think it is very much related and might be fixed here as well.

Good catch, done!

yakutovicha (approved):
Worked for me. Thanks a lot @csadorf!

PR #244, opened 2021-11-15
Improve error message in install with url

unkcpz:
When trying to install the app from a [PEP 508](https://www.python.org/dev/peps/pep-0508/) form with the branch specified, the `@` is wrongly checked and without a print useful message.


csadorf (approved):

csadorf:
Thanks!

PR #245, opened 2021-11-18
Fix "find_installed_packages()" util function for editable installs.

csadorf:
The function would previously fail whenever the were any Python
packages installed in editable mode within the environment.


yakutovicha (approved):
LGTM!

yakutovicha:
> 
> 
> The function previously fails when editable package installations are present in the parsed environment.

I would work a bit on the description, though.

csadorf:
> > The function previously fails when editable package installations are present in the parsed environment.
> 
> I would work a bit on the description, though.

Better?

PR #283, opened 2022-06-01
Fix to order of parameters and image name  in `docker run` line

Crivella:
This PR fixes the following issues: 
 - The ```-p 8888:8888``` should be before the name of the image (otherwise it is passed as a command to the image instead of as a parameter to docker).
 - The default image name when pulling from dockerhub is ```aiidalab/aiidalab-docker-stack```


unkcpz (approved):
@Crivella thanks! all good.

PR #292, opened 2022-07-05
Respect global constraints file when installing app dependencies.

### LINKED ISSUE #220: Avoid overriding packages installed in the system environment.
Currently, even if packages are installed in the base environment, they can be superseded by the ones installed in the `.local` folder. I think that better practice would be to forbid installing the core packages (`aiida-core`, `aiidalab`, `aiidalab-widgets-base`).

csadorf:
Fixes #220.


yakutovicha (approved):
thanks, @csadorf! Looks good to me!

PR #295, opened 2022-07-05
Run post_install script after app installation.

csadorf:


csadorf:
Tested in conjunction with https://github.com/aiidalab/aiidalab-qe/pull/251 .

PR #303, opened 2022-07-28
Better error if logo/categories metadata are missing

danielhollas:
I just lost an hour when trying to understand why the home app was throwing an error about a missing logo key, until I realized that I had an extra folder in the `apps/` folder. With this change I would have seen the issue immediately at least.

### DIFF
--- a/aiidalab/app.py
+++ b/aiidalab/app.py
@@ -585,8 +585,16 @@ class AiidaLabApp(traitlets.HasTraits):
         super().__init__()
 
         self.name = self._app.name
-        self.logo = self._app.metadata["logo"]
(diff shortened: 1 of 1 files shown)


yakutovicha:
@csadorf please have a look, thanks!

csadorf (commented):
It's not 100% clear to me how the error message would help. Was it unclear that the home app was throwing the error?

danielhollas:
@csadorf it was not clear from which app / directory the error was coming from.

csadorf (approved):

csadorf:
@danielhollas Thanks a lot!

PR #315, opened 2022-10-16
Fix: Add missing arguments to raised exception

danielhollas:
Before this change, if a  `post_install` script failed when installing an app we would get
```
raise CalledProcessError
TypeError: __init__() missing 2 required positional arguments: 'returncode' and 'cmd'
```

More info at:
https://docs.python.org/3/library/subprocess.html#subprocess.CalledProcessError


yakutovicha (approved):
Thanks @danielhollas!

PR #323, opened 2022-11-21
Install App dependencies to ~/.local, not /opt/conda

### LINKED ISSUE #322: App dependencies should be installed in user mode
In the old Docker stack image, the conda environment in `/opt/conda` was not writeable by the `aiida` user.
As a side-effect of that, when the AiiDAlab app was installed via pip, its declared dependencies (and the app itself) was installed in the `home/aiida/.local` directory. Thus these dependencies persisted between different container invocation since they were stored in the home volume / directory.

In the new full-stack image, the conda environment is writeable by the `jovyan` user, hence pip default to system-wide installation. Therefore, the installed dependencies disappear when the container exits. In the next invocation, the Apps won't load due to missing dependencies.

The possible fix for this is very simple, just add the `--user` argument pip here:

https://github.com/aiidalab/aiidalab/blob/1093f7e4e9f7ef4e26733e69f5fde5bee71cd7ef/aiidalab/utils.py#L203

@unkcpz I am very surprised this did not come up earlier. Did you not observe this with QeApp?
If you agree with the proposed fix I will prepare a PR.

danielhollas:
In the old Docker stack image, the conda environment in /opt/conda was not writeable by the aiida user. As a side-effect of that, when the AiiDAlab app was installed via pip, its declared dependencies (and the app itself) was installed in the /home/aiida/.local directory. These dependencies were then persisted between different container invocation since they were stored in the home volume.

In the new full-stack image, the conda environment is writeable by the jovyan user, hence pip defaults to the system-wide installation, in this case /opt/conda. Therefore, the installed dependencies disappear when the container exits.
In the next invocation, the Apps won't load due to missing dependencies.

The fix is to explicitly instruct pip to do the user install into the ~/.local/ directory.

Closes #322

**TODO**
- [x] Test that the fix indeed work
- [x] Test the packages in `~/.local/` take precedence before packages in `/opt/conda`

### DIFF
--- a/aiidalab/utils.py
+++ b/aiidalab/utils.py
@@ -200,7 +200,7 @@ def this_or_only_subdir(path):
 
 def run_pip_install(*args, python_bin=sys.executable):
     return subprocess.Popen(
-        [python_bin, "-m", "pip", "install", *args],
+        [python_bin, "-m", "pip", "install", "--user", *args],
         stdout=subprocess.PIPE,
         stderr=subprocess.STDOUT,
     )


danielhollas:
@unkcpz could you please test this fix and the corresponding fix in Dockerfile https://github.com/aiidalab/aiidalab-docker-stack/pull/324 with QeApp. The detailed explanation is given in the issue #322 

If this works, we need to release a new aiidalab version and publish a new docker image.

Could you also perhaps add some permissions for me on this repo so that I can request reviews and stuff like that? Thanks!

CC @yakutovicha

unkcpz:
> Could you also perhaps add some permissions for me on this repo so that I can request reviews and stuff like that?

Sure! done.

I'll take a look and test at this PR.

unkcpz:
Hi @danielhollas, I got the warning `WARNING: The script ansi2html is installed in '/home/jovyan/.local/bin' which is not on PATH.`. To make this change work, I think the path needs to be included in the `PATH`. The problem is where to update `PATH` environment variable, it can be hardcore in bashrc or can update the `PATH`. Meanwhile we need to think the priority of the `$HOME/.local/bin`, I personally think it needs to go before the `/opt/conda/bin` which may to aggressive and will override the system package and therefore break dependencies.

danielhollas:
> Hi @danielhollas, I got the warning WARNING: The script ansi2html is installed in '/home/jovyan/.local/bin' which is not on PATH.. To make this change work, I think the path needs to be included in the PATH.

Good catch. I checked the old AiiDA 1.x Docker image and this folder is indeed in the PATH:

```
/opt/conda/bin:/opt/conda/condabin:/opt/conda/bin:/opt/conda/condabin:/opt/conda/bin:/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin:/home/aiida/.local/bin
```

It seems it was put at the end, so the `/opt/conda/bin` takes precedence,  because In Linux generally user / local paths take precedence before system-wide paths. So this is a bit confusing, but I guess it was done to reduce the possibility of Apps breaking the dependencies.


> The problem is where to update PATH environment variable, it can be hardcore in bashrc or can update the PATH.

I propose to just put this to `.bashrc`, and if we run into issues we can always change it in the future.  This is how it was done in the old image as well.

I've added the change to the PR on the aiidalab-docker-stack. https://github.com/aiidalab/aiidalab-docker-stack/pull/325/files Could you give it a test? I'll try testing it tomorrow as well.

danielhollas:
@unkcpz I believe this one is good to go. Then we can release a new `aiidalab` version and update it on the `aiidalab-docker-stack` repo.

PR #328, opened 2022-11-29
Deps: Bump watchdog

danielhollas:
Watchdog dependency seems to be used for live monitoring of AiiDAlab app installation process (I didn't look too closely what it actually does though). The old version prevents upgrade to Python 3.10.

I checked the watchdog Changelog and it seems this should be a seamless update. I have manually tested installation from the Appstore on this branch and it seems to work. Would be good if somebody else with more experience with Appstore tested as well. You can use this image: `ghcr.io/aiidalab/full-stack:pr-337` Note that this image is build with Python 3.10, so you might as well test other things. :-)

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -41,7 +41,7 @@ install_requires =
     toml~=0.10
     traitlets~=5.0
     urllib3~=1.24
-    watchdog~=0.10
+    watchdog~=2.1
 python_requires = >=3.7
 include_package_data = True
 zip_safe = False


unkcpz (approved):
The bump looks good to me. But I have no experience with this section of code. It also has no tests related. Could be good to run an installation to test it.

danielhollas:
@yakutovicha could you give this a test? This is the only thing that's keeping us from migrating to Python 3.10, so I think we should go for it.

Note that for testing this, if you want to download the image with Python 3.10 from ghcr.io, you will probably need to get a Github access token, that you then pass to Docker to authenticate. Alternatively, you can set the packages as public, I don't really see a downside to that (except possibly confusing people, who generally should always download from Dockerhub).

danielhollas:
Probably should be tested and merged after #346

danielhollas:
@unkcpz I have merged the `isAlive` PR here in case you want to test it together with the Watchdog update. 
I've also rebuild the image that is using this branch, so you can just do

```
docker pull ghcr.io/aiidalab/full-stack:pr-337
```

Images on ghcr.io are not public, so if you've never pulled from it, you need to first create a Github token for docker authentication, details described here: https://docs.github.com/en/packages/working-with-a-github-packages-registry/working-with-the-container-registry#authenticating-to-the-container-registry

unkcpz:
Thanks @danielhollas, I did pull the full-stack from `ghcr.io` once. Is the image will build every time the PR in aiidalab-docker-stack updated?

danielhollas:
Yes, the image is build in the CI automatically for each commit on that PR.

yakutovicha:
Since @unkcpz is on it, I will remove myself from the reviewers. Please proceed without me here 👍

unkcpz:
I give it a test and the issue report in https://github.com/aiidalab/aiidalab/issues/340 did not show. However, the https://aiidalab.github.io is down and the app-registry is broken. I 'll give it another test after @superstar54 fix it.

unkcpz:
I gave it a test again and all works fine. No `isAlive` error popped.

PR #335, opened 2022-12-01
Chore: maintain pre-commit checks.

yakutovicha:


setup.cfg
```diff
@@ -65,9 +65,9 @@ tests =
 
 [flake8]
 ignore =
-    E501  # Line length handled by black.
```
danielhollas, line 68: Should these comments perhaps be moved up instead of deleted? I find them quite useful. But not a blocker...

yakutovicha: I also wanted this, but one of the hooks deletes them 🤷

PR #301, opened 2022-07-27
Do not run reentry scan after app installation

csadorf:

PR #347, opened 2023-02-08
validate the core dependencies

### LINKED ISSUE #318: modify the is_compatible() function too look at the constraints instead of the current environment
After migrating docker stack to support AiiDA 2.x, most apps in app store would incompatible since they are specifically set to dependent on aiida-core~=1.0. We need to disallow the user to install the old versions of these apps until they have aiida-core 2.0 supported.

The `is_compatible()` function will do this check,
https://github.com/aiidalab/aiidalab/blob/04f295b6eb8aee851a4eb3f538369dd441ce3ea9/aiidalab/app.py#L258-L259

unkcpz:
fixes #318

### DIFF
--- a/aiidalab/app.py
+++ b/aiidalab/app.py
@@ -18,6 +18,7 @@ from threading import Thread
 from time import sleep
 from urllib.parse import urldefrag, urlsplit, urlunsplit
 from uuid import uuid4
+from collections import namedtuple
 
 import requests
 import traitlets
@@ -49,6 +50,7 @@ from .utils import (
 
 logger = logging.getLogger(__name__)
 
+Dependency = namedtuple("Dependency", ["installed", "required"])
 
 # A version is usually of type str, but it can also be a value
 # of this Enum to indicate special app states in which the
@@ -257,6 +259,29 @@ class _AiidaLabApp:
 
     def is_compatible(self, version, python_bin=None):
         return not any(self.find_incompatibilities(version, python_bin))
+    
+    def find_dependencies_to_install(self, version_to_install, python_bin=None):
+        """Return a list of dependencies (namedtuple Dependency)."""
+        if python_bin is None:
+            python_bin = sys.executable
+            
+        if not version_to_install:
+            return []
+        
+        unmatched_requirements = [r[1] for r in self.find_incompatibilities(version_to_install, python_bin)]
+        packages = find_installed_packages(python_bin)
+        
+        dependencies_to_install = []
+        for requirement in unmatched_requirements:
+            installed = None
+
+            for package in packages:
+                if requirement.name == package.name:
+                    installed = package
+
+            dependencies_to_install.append(Dependency(installed, requirement))
+
+        return dependencies_to_install
 
     def _install_dependencies(self, python_bin, stdout=None):
         """Try to install the app dependencies with pip (if specified)."""
@@ -585,7 +610,10 @@ class AiidaLabApp(traitlets.HasTraits):
     available_versions = traitlets.List(traitlets.Unicode)
     installed_version = traitlets.Union(
         [traitlets.Unicode(), traitlets.UseEnum(AppVersion)]
-    )
+    )   # installed_version is updated from _AiiDALabApp only
+    version_to_install = traitlets.Unicode(allow_none=True)
+    dependencies_to_install = traitlets.List()
+    strict_dependencies_validation = traitlets.Bool()
     remote_update_status = traitlets.UseEnum(
         AppRemoteUpdateStatus, readonly=True, allow_none=True
     )
@@ -596,6 +624,9 @@ class AiidaLabApp(traitlets.HasTraits):
     detached = traitlets.Bool(readonly=True, allow_none=True)
     compatible = traitlets.Bool(readonly=True, allow_none=True)
     compatibility_info = traitlets.Dict()
+    
+    # packages that need to compatible strictly
+    _CORE_PACKAGES = ["aiida-core"]
 
     def __init__(self, name, app_data, aiidalab_apps_path, watch=True):
         self._app = _AiidaLabApp.from_id(
@@ -647,6 +678,11 @@ class AiidaLabApp(traitlets.HasTraits):
     @traitlets.default("busy")
     def _default_busy(self):  # pylint: disable=no-self-use
         return False
+    
+    @traitlets.observe("version_to_install")
+    def _observe_version_to_install(self, change):
+        if change["old"] != change["new"]:
+            self.refresh()
 
     @contextmanager
     def _show_busy(self):
@@ -720,6 +756,18 @@ class AiidaLabApp(traitlets.HasTraits):
             return not any(incompatibilities)
         except KeyError:
             return None  # compatibility indetermined for given version
+        
+    def _validate_strict_dependencies(self, dependencies: list[Dependency]):
+        """Check core dependencies that are not allowed to be override
+        
+        return True if the strict dependencies satisfied
+        """
+        for dep in dependencies:
+            for package in self._CORE_PACKAGES:
+                if package == dep.required.name:
+                    return False
+
+        return True
 
     def _remote_update_status(self):
         """Determine whether there are updates available.
@@ -735,7 +783,7 @@ class AiidaLabApp(traitlets.HasTraits):
         return False
 
     def _refresh_versions(self):
-        self.installed_version = self._app.installed_version()
+        self.installed_version = self._installed_version()
         self.include_prereleases = self.include_prereleases or (
             isinstance(self.installed_version, str)
             and parse(self.installed_version).is_prerelease
@@ -751,6 +799,10 @@ class AiidaLabApp(traitlets.HasTraits):
                 for version in all_available_versions
                 if self.include_prereleases or not parse(version).is_prerelease
             ]
+            
+    def _refresh_dependencies_to_install(self):
+        version_to_install = self.version_to_install
+        self.dependencies_to_install = self._app.find_dependencies_to_install(version_to_install)
 
     @throttled(calls_per_second=1)
     def refresh(self):
@@ -758,8 +810,13 @@ class AiidaLabApp(traitlets.HasTraits):
         with self._show_busy():
             with self.hold_trait_notifications():
                 self._refresh_versions()
+                self._refresh_dependencies_to_install()
                 self.set_trait(
                     "compatible", self._is_compatible(self.installed_version)
+                )
+                self.set_trait(
+                    "strict_dependencies_validation", 
+                    self._validate_strict_dependencies(self.dependencies_to_install),
                 )
                 self.set_trait(
                     "remote_update_status",


unkcpz:
I am going to test the change for the old docker stack but quickly run into an issue that the `aiidalab` version is strictly pinned in the old stack image https://github.com/aiidalab/aiidalab-docker-stack/blob/dff88d30a86e7f8182ac0390390407441f06c813/requirements.txt#L1. 

The problem is do we really need to backport the changes to current 1.x AiiDAlab productions, the only virtue I can see is after implementing the dynamic check for `Update available` state of versions that may not compatible with core dependencies. Although it is doable by loose the version confinement of aiidalab-docker-stack 22.08.1, I think since this is a new feature rather than a critical bug, it is worthless if we don't go to actively support that in the future.  Any suggestion? @yakutovicha @danielhollas

yakutovicha:
> The problem is do we really need to backport the changes to current 1.x AiiDAlab productions

Let's keep it for the new image only. We have enough burden with other things at the moment.

danielhollas:
Well, given that none of the apps beside the QeApp are migrated yet, I think we really need to backport this at some point. Happy to go forth with this to unblock the new stack for the March deadline, but I think the backport needs to happen before we release AWB 2.0.

@unkcpz for testing, you can simply delete the hardcoded requirement from the `requirements.txt` file, no?

I do not think the backport will be difficult. I've looked through the git history, and it looks like `aiidalab-home` should still be compatible with the old stack. There's one breaking change in `aiidalab` repo, but it only concerns running the `reentry scan`. 

https://github.com/aiidalab/aiidalab/pull/301

It should be possible make it compatible again, simply by detecting the aiida-core version and running `reentry scan` when needed.

Then we could simply release a new version of the old stack with updated `aiidalab-home` and `aiidalab`. But in any case it's better to merge and test thoroughly with the new stack.

tests/test_appclass.py
```diff
@@ -0,0 +1,159 @@
...
+            app_data = EXAMPLE_APP_REGISTRY_ENTRY
+
+        app = AiidaLabApp(name, app_data, aiidalab_apps_path, watch=False)
+
+        return app
+
+    return _generate_app
+
+
+EXAMPLE_APP_REGISTRY_ENTRY = {
```
danielhollas, line 65: Please put this in a separate file (maybe in `static` directory?). It would also look a bit nice in YAML format, but that's up to you, I don't have a strong opinion.

danielhollas (changes requested):
@unkcpz thanks so much for adding tests, that's awesome. :heart: Given how complex the logic is I feel much better with tests.

Will do full careful testing next week, for now just minor comments on the code.

aiidalab/app.py
```diff
@@ -258,6 +262,31 @@ def find_incompatibilities(self, version, python_bin=None):
...
+        for requirement in unmatched_requirements:
+            installed = None
+
+            for package in packages:
+                if requirement.name == package.name:
+                    installed = package
+
+            dependencies_to_install.append(Dependency(installed, requirement))
+
+        return dependencies_to_install
```
yakutovicha, line 288:
```suggestion

        p_dict = {p.name:p for p in packages}
        r_dict = {r.name:r for r in requirements}
        intersection = r_dict.keys() & p_dict.keys()
        return [Dependency(p_dict[i], r_dict[i]) for i in intersection]

```

yakutovicha (changes requested):
Thanks, @unkcpz. Also a bunch of change requests from my side.

aiidalab/app.py
yakutovicha: done in [2c3eab0](https://github.com/aiidalab/aiidalab/pull/347/commits/2c3eab0ad6d006b96a76c40990a0685ead87c2e6)

aiidalab/app.py
```diff
@@ -258,6 +262,31 @@ def find_incompatibilities(self, version, python_bin=None):
     def is_compatible(self, version, python_bin=None):
         return not any(self.find_incompatibilities(version, python_bin))
 
+    def find_dependencies_to_install(self, version_to_install, python_bin=None):
```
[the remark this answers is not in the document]
yakutovicha: > this only returns unmet dependencies that are already installed

I also had this impression from the first look, but it also returns the non-installed packages in the following form:
```python
Dependency(None), Dependency(aiida-core<2.0)
```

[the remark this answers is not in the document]
unkcpz: Yes, if any dependencies in the requirements are not found or found not compatible with the packages in the python binary path, it will be displayed.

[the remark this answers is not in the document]
yakutovicha: I think I made the function better now 😄 [e9beee0](https://github.com/aiidalab/aiidalab/pull/347/commits/e9beee080130bb4cb41a4d15e196a9e4e2bffbe7)

aiidalab/app.py
```diff
@@ -579,24 +608,30 @@ class AiidaLabApp(traitlets.HasTraits):
             If true (default), automatically watch the repository for changes.
     """
 
-    path = traitlets.Unicode(allow_none=True, readonly=True)
+    path = traitlets.Unicode(allow_none=True).tag(readonly=True)
```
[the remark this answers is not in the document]
unkcpz: > same question, why is this needed @unkcpz?

Sorry I forget to reply on this PR. This change is required to suppress the deprecation warning shown in the pytest.
Same as you can find at https://github.com/jupyter-widgets/ipyleaflet/issues/414

> To me, the read-only traits are the ones that should be perceived by the outside world as information providers. They can tell about the status of an object, but can't receive information.

For `AiiDALabApp` class, it wrap the `_AiiDALabApp` dataclass. From my understanding, anything defined in the `_AiiDAlabApp` is the read-only for the traitslet. The variable that interacts with the widget will be r/w.

aiidalab/app.py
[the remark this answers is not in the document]
danielhollas: Cheers @yakutovicha, it's clearer now. Please also update the docstring to better explain what the function does, including the two cases (dependency missing versus dependency wrong version).

aiidalab/app.py
```diff
@@ -579,24 +608,30 @@ class AiidaLabApp(traitlets.HasTraits):
...
 
-    available_versions = traitlets.List(traitlets.Unicode)
+    available_versions = traitlets.List(traitlets.Unicode())
     installed_version = traitlets.Union(
         [traitlets.Unicode(), traitlets.UseEnum(AppVersion)]
-    )
+    )  # installed_version is updated from _AiiDALabApp only
+    version_to_install = traitlets.Unicode(allow_none=True)
+    dependencies_to_install = traitlets.List()
+    strict_dependencies_validation = traitlets.Bool()
```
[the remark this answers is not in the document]
yakutovicha: just did that. Also made a corresponding change in the aiidalab-home

aiidalab/app.py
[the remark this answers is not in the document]
yakutovicha: done in [c38e2d0](https://github.com/aiidalab/aiidalab/pull/347/commits/c38e2d0a25f0da5701b8ce4b1f0c968eb88c0817)

aiidalab/app.py
```diff
@@ -721,6 +761,18 @@ def _is_compatible(self, app_version):
         except KeyError:
             return None  # compatibility indetermined for given version
 
+    def _validate_strict_dependencies(self, dependencies: List[Dependency]):
```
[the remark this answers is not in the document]
yakutovicha: done in [97b0dda](https://github.com/aiidalab/aiidalab/pull/347/commits/97b0ddae107e79bc3528ad56e43138bccc694a3c)

aiidalab/app.py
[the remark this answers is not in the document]
yakutovicha: Let's resolve it then.

aiidalab/app.py
```diff
@@ -49,6 +51,8 @@
 
 logger = logging.getLogger(__name__)
 
+Dependency = namedtuple("Dependency", ["installed", "required"])
```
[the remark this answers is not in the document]
unkcpz: Just come back to this, I think this may not best, if we make this a class and add the method or property `satisfied` which check the core dependencies compatibility directly here will things easier in other places where this class is used. The type hint can also more easily applied to the vairables. The question is whether the `_CORE_PACKAGE` can be defined in the `Dependency`?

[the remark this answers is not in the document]
danielhollas: Indeed, `dataclass` might be a more powerful approach. I would keep the definition of core packages separate though, since that seems logically different.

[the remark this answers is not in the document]
unkcpz: It would be something like:
```python
# class Denpendency is used to store the information of the dependency and check the compatibility of the dependency
# wrt to the core dependencies defined in _CORE_DEPENDENCIES
# the class has two attributes:
# - installed: type of Package which is the package installed for the dependency
# - required: type of Requirement which is the requirement for the dependency
# the property strictly_satisfied return True if compatible with the core dependencies
class Dependency:
    
    _CORE_DEPENDENCIES = ["aiida-core"]
    def __init__(self, installed: Package, required: Requirement):
        self.installed = installed
        self.required = required

    @property
    def strictly_satisfied(self):
        # check the installed version wrt to the _CORE_DEPENDENCIES
        if self.installed.name in self._CORE_DEPENDENCIES:
            return self.installed.version in self.required
        else:
            return self.installed.version in self.required or self.installed.version > self.required
```

I need to go now, let's discuss tomorrow and see if we want to go far in this.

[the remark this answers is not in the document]
danielhollas: Nice. I think with dataclass decorator you should be able to simplify like this

```python
@dataclass(frozen=True)
class Dependency:
    _CORE_DEPENDENCIES = ["aiida-core"]
    installed: str
    required: str

    @property
    def strictly_satisfied(self):
.
.
.
```

tests/test_appclass.py
```diff
@@ -0,0 +1,159 @@
...
+    app.version_to_install = "v22.11.0"
+    assert app.strict_dependencies_validation is False
+
+    app.version_to_install = "v22.11.1"
+    assert app.strict_dependencies_validation is True
+
+
+# The data for test purpose as fixture
+@pytest.fixture
+def generate_app():
```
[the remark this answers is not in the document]
yakutovicha: done in 2ed9d6f

tests/test_appclass.py
yakutovicha: done in 2ed9d6f

aiidalab/app.py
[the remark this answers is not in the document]
danielhollas: @yakutovicha just pointing out this discussion above since you're now working on this.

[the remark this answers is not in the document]
yakutovicha: I ended up replacing the whole class with a dictionary `{"installed":.., "required":..}` in [65f83cd](https://github.com/aiidalab/aiidalab/pull/347/commits/65f83cde6eed1d601c3ad39a4f428f728f9c0e9c). I don't think there is a reason to keep it since the versions with non-matching core dependencies are not shown.

yakutovicha:
ready for testing @unkcpz and @danielhollas

aiidalab/app.py
```diff
@@ -648,6 +699,20 @@ def _default_detached(self):
     def _default_busy(self):  # pylint: disable=no-self-use
         return False
 
+    @traitlets.validate("version_to_install")
+    def _validate_version_to_install(self, proposal):
```
unkcpz, line 703: This function is not used anywhere, correct?

aiidalab/app.py
```diff
@@ -228,12 +235,29 @@ def find_matching_releases(self, specifier, prereleases=None):
...
+                return False
+        return True
+
     @staticmethod
     def _find_incompatibilities_python(requirements, python_bin):
         packages = find_installed_packages(python_bin)
         for requirement in map(Requirement, requirements):
-            f = [p for p in packages if p.fulfills(requirement)]
-            if not any(f):
+            pkg = packages.get(requirement.name)
```
unkcpz, line 257:
```suggestion
            pkg = packages.get(requirement.name, None)
```

packages is a dict so the `get` need has a default return.

unkcpz (commented):
@yakutovicha, just some minor requests. Looks in very good shape now.

aiidalab/app.py
yakutovicha: It is called every time one wants to update the trait. And it does validate the trait. See in the tests:

```python
    # The version `v22.11.0` is incompatible while `v22.11.1` is compatible
    with pytest.raises(traitlets.TraitError):
        app.version_to_install = "v22.11.0"
    app.version_to_install = "v22.11.1"
```

aiidalab/app.py
danielhollas: `None` is the default return value for `dict.get()` so no need to specify it I think

yakutovicha: > `None` is the default return value for `dict.get()` so no need to specify it I think

that's correct.

tests/test_appclass.py
```diff
@@ -0,0 +1,40 @@
...
+    app = generate_app()
+
+    # without prereleases tick
+    app.refresh()
+    assert app.has_prereleases
+    assert app.include_prereleases is False
+
+    # tick prereleases tick
+    app.include_prereleases = True
+    assert "v23.01.0b1" in app.available_versions
```
danielhollas, line 27: The negative of this assert should be few lines above before the `include_prereleases` is set to True

aiidalab/app.py
danielhollas: Nice solution, thanks! :+1:

aiidalab/app.py
```diff
@@ -157,9 +158,15 @@ def installed_version(self):
             return self.metadata.get("version", AppVersion.UNKNOWN)
         return AppVersion.NOT_INSTALLED
 
-    def available_versions(self):
+    def available_versions(self, python_bin=None):
+        """Return a list of available versions excluding the ones with core dependency conflicts."""
         if self.is_registered:
-            yield from sorted(self.releases, key=parse, reverse=True)
+            for version in sorted(self.releases, key=parse, reverse=True):
+                version_requirements = self.releases[version].get("environment")[
```
danielhollas, line 165: You're using get here but not checking for None. Is it safe to assume that 'environment' and 'python_requirements` are always in the app registry for a given release? Is it robust if an app doesn't define any python requirements?

danielhollas (changes requested):
This looks great, thanks @yakutovicha! Just one question and one request for the tests. I will test this next week.

aiidalab/app.py
unkcpz: Interesting! Thanks for clearing up.

aiidalab/app.py
unkcpz: I didn't know that 😿, thanks both!

aiidalab/app.py
unkcpz: I think if the python requirements are not defined the `version_requirements` will be `None` and it is okay to pass the `None` to function `_strict_dependencies_met`. That simply means there are not dependencies required and the strict dependencies will meet anyway.

tests/test_appclass.py
unkcpz: @danielhollas You are asking for the changes or it is a comment we can add?

unkcpz (commented):
Hi @danielhollas, thanks for reviewing. Let me know if you think it is ready for to me merged.

tests/test_appclass.py
danielhollas: I am asking to add an extra assert above line 24 which will test that the pre-release version is not included when the "include_prerelease" is not ticked.

```python
assert "v23.01.0b1 not in app.available_versions
```

aiidalab/app.py
danielhollas: @unkcpz  I still think this code is backwards. Here's what will happen with the current code if

 - If the `environment` key is not there, the get() call will produce `None["python_requirements"]` which will raise.
 - If `python_requirements` are missing, we raise again since we're not using get()
 
I think the code should instead be (if we're assuming that the `environment` key is always there).
 ```suggestion
                version_requirements = self.releases[version]["environment"].get("python_requirements")
```

Please actually test this by modifying the mocked registry file in the test suite (perhaps add another entry with empty python requirements). Looking at the `strict_dependencies_met` function, I think if you pass it None it will raise as well since it is iterating over the requirements without first checking if they are None.

File: aiidalab/app.py [hunk 13]
```diff
@@ -228,12 +235,29 @@ def find_matching_releases(self, specifier, prereleases=None):
         matching_releases.sort(key=parse, reverse=True)
         return matching_releases
 
+    @staticmethod
+    def _strict_dependencies_met(requirements, python_bin):
+        """Check whether the given requirements are compatible with the core dependencies of a package."""
```
danielhollas, line 240:
```suggestion
        """Check whether the given requirements are compatible with the core dependencies of a package."""
            if requirements is None or len(requirements) == 0:
                return True 
```

aiidalab/app.py
unkcpz: I tried but I don't think it is a good idea to create a "fake" mocked registry entity. I don't know if the aiidalab registry already handle the case if there is nothing in the requirement and in that case the `environment` will be an empty dict or will just not appeared in the registry file. 
Instead of adding a mock test, I give the default value for every attribute get. by `.get("environment", {}).get("python_requirments", [])`. I think this will make things more safe even we don't have unittest for that.

aiidalab/app.py
unkcpz: This also covered by the default value empty list `[]` returned no dependencies defined in the file.

unkcpz: But in this case, I realised that using the python type hint will prevent lots of cases where these issues may happend. I'll open an issue to bring the mypy into these packages, at least for `aiidalab` and `aiidalab-home` first.

aiidalab/app.py
danielhollas: Fair enough, but I think we should be proactive here and actually find out what happens when an app doesn't declare any dependencies, because that's something that can very well happen. Will it crash the registry or will it be fine? What would the resulting json file look like? Once we know that, I would add that test case here, there's nothing fake about it.

danielhollas:
I need more time to test this properly. I am away the whole tomorrow so will pick this up on Wednesday.

Here's one strange thing I noticed though. I have an AWB installed in an editable mode in my full-stack environment (with AiiDA 2.2). In the app, I get a warning, as I should, that the latest stable release of AWB, 1.4.2 is not compatible. However, here's what I get for the unmet dependencies.

image

It says that `jupyter-client` and `vapory` are not installed, but in fact they are installed already.

```console
$ pip show vapory
ame: Vapory
Version: 0.1.2
Summary: 3D rendering with Python and POV-Ray
Home-page: UNKNOWN
Author: Zulko
Author-email: 
License: see LICENSE.txt
Location: /home/jovyan/.local/lib/python3.9/site-packages
Requires: 
Required-by: aiidalab-widgets-base

$ pip show jupyter-client
Name: jupyter_client
Version: 7.3.5
Summary: Jupyter protocol implementation and client libraries
Home-page: 
Author: 
Location: /opt/conda/lib/python3.9/site-packages
Requires: entrypoints, jupyter-core, nest-asyncio, python-dateutil, pyzmq, tornado, traitlets
Required-by: ipykernel, jupyter-server, nbclassic, nbclient, notebook
```

@unkcpz can you take a look? How is the list of currently installed packages determined at the moment?

danielhollas:
Oh, looking at the screenshot above, I would change the UI a bit. For the installed packages on the left, they are wrapped in `Package()`. That's confusing to me. I would expect it to look more like this

```
The following dependencies will be changed
  - ase==3.22.1 --> ase<3.20
```

danielhollas:
One more thing. I have an editable install of AWB, checked out at the lastest commit on master branch. This is obviously compatible with my aiida-core.

However, I still see the warning that says "The installed version of this app is not compatible with this AiiDAlab environment."
So I guess there is some mismatch and the current code does not take into account locally modified installation, where it should parse the setup.cfg from them (instead of taking info from the latest stable version from the registry)

image

aiidalab/app.py
```diff
@@ -49,6 +49,8 @@
 
 logger = logging.getLogger(__name__)
 
+_CORE_PACKAGES = ["aiida-core"]
```
danielhollas, line 52: I noticed that some applications declare dependency on `aiidalab`, e.g.
https://github.com/epfl-theos/aiidalab-aurora/blob/develop/setup.cfg#L31

I would propose to add `aiidalab` and `aiidalab-home` amongst the core dependencies to avoid the user to re-install them.

danielhollas (changes requested):

aiidalab/app.py
danielhollas: Ugh, this is more annoying than I expected. Unfortunately, for some reason AWB versions >1.3.4 allow aiida-core 2.0

https://github.com/aiidalab/aiidalab-widgets-base/blob/support/aiida-1.x/setup.cfg#L22

I would also add a bit of the jupyter stack to the core dependencies, because downgrading these packages for sure breaks the whole enviroment. Specifically, AWB specifies dependencies on `nbconvert` and `jupyter-client` so I would add at least these two, which would hide these old AWB versions.

unkcpz: Agree with this, I think `jupyter-client`, `aiidalab` and `aiidalab-home` is enough to pin most of the dependencies.

unkcpz:
> Oh, looking at the screenshot above, I would change the UI a bit. For the installed packages on the left, they are wrapped in `Package()`. That's confusing to me. I would expect it to look more like this
> 
> ```
> The following dependencies will be changed
>   - ase==3.22.1 --> ase<3.20
> ```

Addressed by https://github.com/aiidalab/aiidalab/pull/347/commits/99d1f6149dc02d1000d11b4a3a8a859a91bef2c7

unkcpz:
> I need more time to test this properly. I am away the whole tomorrow so will pick this up on Wednesday.
> 
> Here's one strange thing I noticed though. I have an AWB installed in an editable mode in my full-stack environment (with AiiDA 2.2). In the app, I get a warning, as I should, that the latest stable release of AWB, 1.4.2 is not compatible. However, here's what I get for the unmet dependencies.
> 
> image
> 
> It says that `jupyter-client` and `vapory` are not installed, but in fact they are installed already.
> 
> ```
> $ pip show vapory
> ame: Vapory
> Version: 0.1.2
> Summary: 3D rendering with Python and POV-Ray
> Home-page: UNKNOWN
> Author: Zulko
> Author-email: 
> License: see LICENSE.txt
> Location: /home/jovyan/.local/lib/python3.9/site-packages
> Requires: 
> Required-by: aiidalab-widgets-base
> 
> $ pip show jupyter-client
> Name: jupyter_client
> Version: 7.3.5
> Summary: Jupyter protocol implementation and client libraries
> Home-page: 
> Author: 
> Location: /opt/conda/lib/python3.9/site-packages
> Requires: entrypoints, jupyter-core, nest-asyncio, python-dateutil, pyzmq, tornado, traitlets
> Required-by: ipykernel, jupyter-server, nbclassic, nbclient, notebook
> ```
> 
> @unkcpz can you take a look? How is the list of currently installed packages determined at the moment?

This is a very important catch!! Thanks a lot for scrutinizing the details for every app. The issue is addressed by https://github.com/aiidalab/aiidalab/pull/347/commits/e9d3ce3a4d0458c2df053bcb5f93e562bf09cdb0 which is very subtle and I think also related to issue report long ago in https://github.com/aiidalab/aiidalab-widgets-base/issues/402.

aiidalab/app.py
unkcpz: I add `jupyter-client` at the moment to temporarily resolve the  `aiidalab-widgets-base` issue you mentioned here.

aiidalab/app.py
unkcpz: I'll try the mock test, since I find the app registry generator code also here. I think we can test them together from a mocked setup.cfg content.

unkcpz: I have a close look at the registry tool, it is safe to get both `environment` and `python_requirement`.
For `environment` it is the key defined in https://github.com/aiidalab/aiidalab/blob/3e8b43135f605cb2af3af25bbbd131da815d0ab9/aiidalab/utils.py#L101-L104.
For `python_requirement`, it is the key mandatory exist in https://github.com/aiidalab/aiidalab/blob/3e8b43135f605cb2af3af25bbbd131da815d0ab9/aiidalab/environment.py#L59-L68

unkcpz: I will not work on filling the test gap at the moment. I'll open an issue for doing it in the future.

aiidalab/utils.py
```diff
@@ -151,9 +152,12 @@ def __init__(self, name, version):
         self.name = name
```
danielhollas, line 152: Also per the comment below, I think it makes sense to canonicalize the name here, otherwise we'll keep hitting this issue.
```suggestion
        self.name = canonicalize_name(name)
```

aiidalab/app.py
```diff
@@ -228,12 +238,29 @@ def find_matching_releases(self, specifier, prereleases=None):
         matching_releases.sort(key=parse, reverse=True)
         return matching_releases
 
+    @staticmethod
+    def _strict_dependencies_met(requirements, python_bin):
+        """Check whether the given requirements are compatible with the core dependencies of a package."""
+        packages = find_installed_packages(python_bin)
```
danielhollas, line 244: Please make sure we use canonical names all throughout this function.

danielhollas (changes requested):
@unkcpz thank you for the fix. I no longer see the spurious Vapory warning. :tada: 

However, the change seems to have broken the core dependency check. Even though `jupyter-client` is included in core dependencies, I still the old version of AWB in the available versions.

image

We need to make sure we use the canonical names everywhere. 

This issue should have been caught by the tests, but clearly wasn't. Please, add some test that would cover packages where canonical names differ from the package name. Thanks!
Also please add a test that will verify that the `_CORE_PACKAGES` contain canonicalized names.

aiidalab/app.py
danielhollas: > In principle the aiidalab and aiidalab-home is only maintained by core developers but not app developers.

The core dependencies serve as a safeguard against applications that would break the environment. It doesn't really matter what app is maintained by whom, we have to assume the worst, and as I provided link above, some apps already specify `aiidalab` in their setup.cfg. To me, both `aiidalab` and `aiidalab-home` are core packages tied to the released Docker stack image and they should never be reinstalled. If users need to upgrade them, they should download the new image. Developers should use the `aiidalab-dev` to allow for easy development.

For AWB, we cannot really constrain it, since different apps may need different versions, and need to be able to reinstall it if needed. 

For `ipywidgets`, I would say that is tied to the AWB, and hence we cannot constrain if we want to be able to update AWB package.

My suggestion to be on the safe side is to add packages that are inherent to the Docker stack, `aiidalab` and `aiidalab-home`. But if you disagree let's open an issue and discuss more, no need to solve it here. I think this also ties to the discussion about the definition of AiiDAlab release, and about AWB.

unkcpz:
Hi @danielhollas, I add some very basic unit tests. I have to admit it is a bit painful to try this. Lots of functions coupled to each other. We for sure need some time to refactor this a bit by adding more unit tests.

aiidalab/app.py
unkcpz: Yes, I did this in new commit.

aiidalab/utils.py
```diff
@@ -172,10 +176,27 @@ def find_installed_packages(python_bin=None):
...
+    """Return the package with the given requirment name from the dict of packages.
+    The usual get method of dict does not work here, because the requirement name and the package name
+    may be different by the case of whether being canonicalized or not.
+
+    For example, the requirement name is 'jupyter-client' and the package name is 'jupyter_client'.
+    The implementation of this method is inspired by https://github.com/pypa/pip/pull/8054
+    """
+    for package in packages.values():
+        # print(package.name, package.version)
+        if canonicalize_name(package.name) == canonicalize_name(name):
```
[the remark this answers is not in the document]
unkcpz: I make `canonicalized_name` a property of Package objcet and reuse it.

danielhollas (changes requested):
@unkcpz thanks so much for adding the tests! Great start. Also thank you for adding type annotations. Could you please also add the type annotations for other functions that are critical in this PR? Also, please add the return type everywhere, that will make things much easier to follow.

My main general comment is this. Now that we're down this rabbit hole of canonical names, I think we need to do it properly and consistently everywhere. I would propose, that each function that is concerned with packages / requirements never receives or returns a string, it should always work with either Requirement or Packages types, or their collections (list or dict). Otherwise, we will never untangle this mess I think and bugs will always be lurking.

unkcpz:
> I would propose, that each function that is concerned with packages / requirements never receives or returns a string, it should always work with either Requirement or Packages types, or their collections (list or dict). Otherwise, we will never untangle this mess I think and bugs will always be lurking.

I second this. I was planning to do this but was very tired after figuring out the monkeypatch stuff. I'll give it a try late tomorrow, need to prepare some presentation in the morning. Thanks for bringing this up, to be honest sometime I can be a bit lazy to stop at low-quality code even though I know how to improve it. Good push 😄 !

aiidalab/app.py
```diff
@@ -228,12 +240,34 @@ def find_matching_releases(self, specifier, prereleases=None):
         matching_releases.sort(key=parse, reverse=True)
         return matching_releases
 
+    @staticmethod
+    def _strict_dependencies_met(requirements: list[str], python_bin):
+        """Check whether the given requirements are compatible with the core dependencies of a package."""
+        from aiidalab.utils import find_installed_packages
```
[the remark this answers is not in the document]
unkcpz: Yes, because of test, I don't know how to make test work without doing this. Since monkeypatch can only patch object, the `find_installed_packages` import in the file head cannot be overridden. If you know how to make this work it would be great. Here is some information, but I find none of the solution work. 

[1] https://stackoverflow.com/questions/31306080/pytest-monkeypatch-isnt-working-on-imported-function
[2] https://stackoverflow.com/questions/70083364/pytest-monkeypatch-doesnt-apply-to-imported-function

unkcpz:
Hi @danielhollas, please have a look at the last commit https://github.com/aiidalab/aiidalab/pull/347/commits/abe2a1412404eda98d41b7fd46f53c567f88ca0b. Instead of still using the `get_package_from_requirement`, I change it to `get_package_from_name` so it is more general since it also used for get the package from the core package list defined. 
Other changes are applied, when it comes to the requirement, the `Requirement` object is always passed to avoid the ambiguity. The tests added all pass, which makes me more confident.

aiidalab/app.py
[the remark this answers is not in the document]
danielhollas: I see. Looking at the `find_installed_packages`, what we really want to monkeypatch is the output from `pip list` command. Therefore, I would propose to make a smaller function called `run_pip_list` or something like that, which you then call inside the `find_installed_packages`, and can monkeypatch in tests (maybe?).

aiidalab/utils.py
```diff
@@ -165,6 +167,9 @@ def canonical_name(self):
 
     def fulfills(self, requirement):
         """Returns True if this entry fullfills the requirement."""
+        if self.version is None:
```
danielhollas, line 170: Why is this here? I think this is not correct, because you need to first check that the names match.

aiidalab/app.py
```diff
@@ -157,9 +162,20 @@ def installed_version(self):
             return self.metadata.get("version", AppVersion.UNKNOWN)
         return AppVersion.NOT_INSTALLED
 
-    def available_versions(self):
+    def available_versions(self, python_bin=None):
```
danielhollas, line 165: Please add type annotations. Are we returning a Package?

aiidalab/app.py
```diff
@@ -228,12 +244,38 @@ def find_matching_releases(self, specifier, prereleases=None):
...
+                and installed_core_package is not None
+                and not installed_core_package.fulfills(
+                    requirements_dict[core.canonical_name]
+                )
+            ):
+                return False
+        return True
+
     @staticmethod
     def _find_incompatibilities_python(requirements, python_bin):
```
danielhollas, line 270: Type annotations please

aiidalab/app.py
```diff
@@ -258,6 +300,36 @@ def find_incompatibilities(self, version, python_bin=None):
     def is_compatible(self, version, python_bin=None):
         return not any(self.find_incompatibilities(version, python_bin))
 
+    def find_dependencies_to_install(self, version_to_install, python_bin=None):
```
danielhollas, line 303: Types

aiidalab/app.py
```diff
@@ -228,12 +244,38 @@ def find_matching_releases(self, specifier, prereleases=None):
...
         for requirement in map(Requirement, requirements):
-            f = [p for p in packages if p.fulfills(requirement)]
-            if not any(f):
+            pkg = get_package_by_name(packages, requirement.name)
+            if pkg is None:
+                yield requirement
+            elif not pkg.fulfills(requirement):
                 yield requirement
 
     def find_incompatibilities(self, version, python_bin=None):
```
danielhollas, line 281: Types

tests/test_appdataclass.py
```diff
@@ -0,0 +1,103 @@
...
+@pytest.fixture
+def python_bin():
+    """Return the path to the python executable."""
+    return sys.executable
+
+
+def test_strict_dependencies_met_default(monkeypatch, python_bin):
+    """Test method _strict_dependencies_met of _AiidaLabApp.
+    Checking the requirements of the app against the core packages."""
+    monkeypatch.setattr(
```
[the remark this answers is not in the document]
danielhollas: Please move this to  a fixture, something like

```python
@pytest.fixture
def monkeypatch_installed_packages(monkeypatch):
    monkeypatch.setattr(
        "aiidalab.utils.find_installed_packages",
        lambda _: _MONKEYPATCHED_INSTALLED_PACKAGES,
    )
```

danielhollas (changes requested):
@unkcpz thanks so much for the changes, looks much better now. My final asks:

1. Please make a separate fixture for monkeypatching the installed packages
2. Add typing to all the functions that relate to packages / requirements. This will make us double check that everything is aligned.

Before you ask me for the final review, please do a manual test that everything works.

unkcpz:
> . Therefore, I would propose to make a smaller function called run_pip_list or something like that, which you then call inside the find_installed_packages, and can monkeypatch in tests (maybe?).

This works! Thanks @danielhollas

aiidalab/utils.py
unkcpz: You are correct, I change the check to the return clause:

```
        return (
            self.canonical_name == canonicalize_name(requirement.name)
            and (self.version in requirement.specifier or self.version is None)
        )
```

aiidalab/app.py
```diff
@@ -752,12 +839,18 @@ def _refresh_versions(self):
                 if self.include_prereleases or not parse(version).is_prerelease
             ]
 
-    @throttled(calls_per_second=1)
+    def _refresh_dependencies_to_install(self):
+        self.dependencies_to_install = self._app.find_dependencies_to_install(
+            self.version_to_install
+        )
+
+    @throttled(interval=10)
```
unkcpz, line 847: This change is important. Otherwise every 1s the app will refresh which is too short. I don't know why this feature is not working. But the blink of the app in every 1s seems like a bug but actually the expected behavior.

aiidalab/app.py
```diff
@@ -752,12 +839,18 @@ def _refresh_versions(self):
                 if self.include_prereleases or not parse(version).is_prerelease
             ]
 
-    @throttled(calls_per_second=1)
+    def _refresh_dependencies_to_install(self):
+        self.dependencies_to_install = self._app.find_dependencies_to_install(
+            self.version_to_install
+        )
+
+    @throttled(interval=1)
```
unkcpz, line 847: I think this can be 10, means the app manager update every 10s.

aiidalab/app.py
```diff
@@ -648,6 +717,24 @@ def _default_detached(self):
     def _default_busy(self):  # pylint: disable=no-self-use
         return False
 
+    @traitlets.validate("version_to_install")
+    def _validate_version_to_install(self, proposal):
```
unkcpz, line 721: There is a race condition for this traitlets. The `self.available_version` is updated in `refresh()` which is triggered by `_observe_version_to_install`. This makes unit tests fail.

unkcpz (commented):
I find a problem of `@traitlets.validate("version_to_install")`, which lead to the fali test. But I don't know how to fix it. The order of observe and validate of traitlets has mess up order.

aiidalab/app.py
unkcpz: This is an old issue if you checkout to the main branch. You can find the app that is already installed. It will keep on flashing in the 1s interval, which is very annoying.

I can open a separate PR for this. What you think? @danielhollas @yakutovicha

danielhollas: Yes, let's not complicate this PR more. Please revert these changes and create an issue.

aiidalab/utils.py
```diff
@@ -147,23 +148,43 @@ def wrapped(instance, *args, **kwargs):
 class Package:
     """Helper class to check whether a given package fulfills a requirement."""
 
-    def __init__(self, name, version):
-        self.name = name
+    def __init__(self, name: str, version: str = None):
```
danielhollas, line 151: Why was this None default added? Can you explain the use case?

aiidalab/utils.py
```diff
@@ -172,10 +193,22 @@ def find_installed_packages(python_bin=None):
         encoding="utf-8",
         capture_output=True,
     ).stdout
-    return [
-        Package(name=package["name"], version=package["version"])
-        for package in json.loads(output)
-    ]
+
+    return output
```
danielhollas, line 197: It makes more sense to me if this method already returned the parsed json object.

```suggestion
    return json.loads(output)
```

danielhollas (changes requested):
Just two minor comments before I test this.
@unkcpz can you merge main into this branch?

I am not sure how to help with the failing test because I do not fully understand what is wrong.

aiidalab/utils.py
unkcpz: This is used in `_CORE_PACKAGES = [Package("aiida-core"), Package("jupyter-client")]`, as you suggest, I use the Package instance and we shouldn't specify version to them.

aiidalab/utils.py
unkcpz: I second it, changed.

aiidalab/utils.py
danielhollas: Oh, I see, got it. Thanks!

danielhollas:
@unkcpz tests are passing :tada: Looks like everything is ready here. Could you please merge in `main` and let's test this one last time before merge.

unkcpz:
> Could you please merge in main and let's test this one last time before merge.

Okay, done.

unkcpz:
I also want to mention that after revert the interval refresh change, for the app installed there is an annoying issue it will keep updating the status. It is not introduced here. I open issue #358 and solution #359 .

danielhollas (approved):
@unkcpz when testing in the old Docker stack, I noticed an issue with the Update button, I've described details in #360. I'd propose to merge this anyway and tackle that issue ASAP in a subsequent PR.

unkcpz:
Cheers! @yakutovicha @danielhollas 🚀  
I hope the commit message is clear (anyway, not able to change 😛).

yakutovicha:
> I hope the commit message is clear (anyway, not able to change 😛).

it is really well written 👍

Issue #334, opened 2022-12-01
aiidalab does not detect interrupted app installation


danielhollas:
If I trigger the installation of an app and interrupt it during the pip installation of dependencies,
aiidalab still subsequently says that the app is installed. I guess we need to have some kind of lockfile to make sure that installation finishes successfully?

Steps to reproduce: 

1. Launch installation with `aiidalab install`
2. Interrupt the installation in the middle
3. `aiidalab list` will show application as installed

danielhollas:
Comment explaining where the issue could be resolved. 

https://github.com/aiidalab/aiidalab/pull/356#discussion_r1126309310

PR #349, opened 2023-02-08
Using INSTALLED_FLAG for installed check

unkcpz:
This is the root of the issue https://github.com/aiidalab/aiidalab-home/issues/100.
The `is_installed` check if the app is installed by checking the folder under `apps`. It is not enough since during the installing, the repo will be put in the path which make the `is_installed` fakely manifest that the app is installed. 
I implement a more strict check by putting a flag file in the app path. The flag is added only if the installed is complete without exception.

- [x] Problem that the Uninstalled button behavior is incorrect. When the app is installed the button is not activated.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -10,3 +10,4 @@ Pipfile.lock
 requirements.txt_dev
 requirements.txt_stable
 .DS_Store
+.INSTALLED.FLAG
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -35,7 +35,7 @@ repos:
             args: [--count, --show-source, --statistics]
 
     - repo: https://github.com/pycqa/isort
-      rev: 5.11.4
+      rev: 5.12.0
       hooks:
           - id: isort
             args: [--profile, black, --filter-files]
--- a/aiidalab/app.py
+++ b/aiidalab/app.py
@@ -75,6 +75,8 @@ class _AiidaLabApp:
     path: Path
     releases: dict = field(default_factory=dict)
 
+    _INSTALLED_FLAG = "INSTALLED.FLAG"
+
     @classmethod
     def from_registry_entry(cls, path, registry_entry):
         return cls(
@@ -166,8 +168,10 @@ class _AiidaLabApp:
             return self._repo.dirty()
 
     def is_installed(self):
-        """The app is installed if the corresponding folder is present."""
-        return self.path.exists()
+        """The app is installed if the corresponding flag file is present."""
+        installed_flag = self.path / self._INSTALLED_FLAG
+
+        return installed_flag.exists()
 
     def remote_update_status(self, prereleases=False):
         """Determine the remote update satus.
@@ -437,6 +441,9 @@ class _AiidaLabApp:
                     f"Failed to install '{self.name}' (version={version}) at '{self.path}'"
                     f", due to error: {error}"
                 )
+
+        else:
+            self.path.touch(self._INSTALLED_FLAG)
 
 
 class AppNotInstalledException(Exception):
--- a/aiidalab/app.py
+++ b/aiidalab/app.py
@@ -49,6 +49,8 @@ from .utils import (
 
 logger = logging.getLogger(__name__)
 
+_INSTALLED_FLAG = ".INSTALLED.FLAG"
+
 
 # A version is usually of type str, but it can also be a value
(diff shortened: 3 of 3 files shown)


unkcpz:
However, this brings the problem that for the already installed app, there is no flag file in the path. I think we need deal with it along with the DB migration, by providing a tool for users to conduct migration by them to apply changes we ask them to do.

danielhollas:
This is great, thanks @unkcpz! We've just run into a problem of incomplete installation with one of the students. I am glad this will get resolved. Note that this also pertains to issue #334.

> However, this brings the problem that for the already installed app, there is no flag file in the path. I think we need deal with it along with the DB migration, by providing a tool for users to conduct migration by them to apply changes we ask them to do.

To make this backwards compatible, I would propose to invert the logic. Moreover, I think we need to be able to distinguish between an installation in progress, and failed installation. Instead of INSTALLED_FLAG, I'd propose

1. Create a before the start of the installation. Remove it at the end, regardless whether the installation failed or not.
2. I am not sure what happens when installation fails. Do we remove the app folder? If not, I'd propose to create something like "INSTALLATION_FAILED" file in the app folder. This would then be an indicator of failed install.

To test this, besides the widget status during install, one needs to test what happens when the installation is interrupted, as described in #334. That's easier to do in command line, by running `aiidalab install` and interrupting it. We than need to have a logic to both detect lockfiles, and failed installs.

danielhollas:
@yakutovicha that's a good point, but I don't now how we would reverse the order, since we currently do

1. Clone git repo
2. run `pip install .` from the repo

yakutovicha:
> @yakutovicha that's a good point, but I don't now how we would reverse the order, since we currently do
> 
> 1. Clone git repo
> 2. run `pip install .` from the repo

I would put it to a sandbox folder first, install the deps and only then move the app to the `apps` folder.

danielhollas:
That sounds good! I think having a lockfile is a crucial first step to make the other things work, and sandboxing could be done as a follow up. Could you create an issue so we don't forget?

yakutovicha:
> That sounds good! I think having a lockfile is a crucial first step to make the other things work, and sandboxing could be done as a follow up. Could you create an issue so we don't forget?

To be honest, I don't think that adding the flag is a good way to go. What are the benefits?

Moreover, looking at the code, I think 99% of things are implemented.

We need to pull the folder to a temporary location (already implemented) and install dependencies(already implemented). If that was successful - copy the folder to `~/apps`. It is really about reshuffling the code a tiny bit.

Happy to make a concurrent PR so we can have a look and pick up a better option. @danielhollas, @unkcpz what do you guys think?

unkcpz:
Thanks @yakutovicha. I like your suggestion, to have a sandbox folder for installation. As you said, all the code is already in the repo feel free to make a concurrent PR.

unkcpz:
Maybe you can have a look at https://github.com/aiidalab/aiidalab/pull/347 first, there are basic tests added. I think that will be valuable for the new implementation.

danielhollas:
> To be honest, I don't think that adding the flag is a good way to go. What are the benefits?
> We need to pull the folder to a temporary location (already implemented) and install dependencies(already implemented). If that was successful - copy the folder to ~/apps. It is really about reshuffling the code a tiny bit.

Fair point, I guess sandboxing should take care of most of the issues, if it is implemented in a way where you can detect installation in progress.  Otherwise user can might trigger the installation twice by mistake (for example by refreshing the page and hitting install again. But maybe this is already take care of. 

> Happy to make a concurrent PR so we can have a look and pick up a better option. @danielhollas, @unkcpz what do you guys think?

:+1:

yakutovicha:
Thanks to both for the quick feedback, really appreciated 👍 

> Maybe you can have a look at #347 first, there are basic tests added. I think that will be valuable for the new implementation.

Let's then first finalize the #347 and then I make a concurrent PR.

yakutovicha:
I put this PR `on hold` (just added a new label).

danielhollas:
> (just added a new label).

@yakutovicha  Off topic, but I'd think the existing 'blocked' would be enough, in the interest of trying to keep the number of labels low.

danielhollas:
I've opened a new issue to track this, #364. I looked the installation code a bit more, and I completely agree what Sasha writes above, sandboxing approach and reversing the order of installation is a way to go.  I'll close this for now.

PR #355, opened 2023-02-24
Make the package compatible with old docker stack again

### LINKED ISSUE #348: Backport validation of core dependencies to old docker stack
As mentioned in #347, the backport support for old stack is required.

danielhollas:
This PR essentially reverts some of the changes made in #301, which dropped support for the old Docker stack.
We reintroduce support by:

1. Detecting `aiida-core` version and running `reentry scan` only for AiiDA 1.x
2. Dropping dependency on `aiida-core` in setup.cfg. I don't really see a reason why this package should depend on aiida-core. This also makes local development easier since one does not need to have aiida-core in the dev environment. Also makes installation in our CI faster.

Closes #348

### DIFF
--- a/aiidalab/app.py
+++ b/aiidalab/app.py
@@ -41,6 +41,7 @@ from aiidalab.utils import (
     load_app_registry_index,
     run_pip_install,
     run_post_install_script,
+    run_reentry_scan,
     run_verdi_daemon_restart,
     split_git_url,
     this_or_only_subdir,
@@ -340,6 +341,11 @@ class _AiidaLabApp:
             process.wait()
             if process.returncode != 0:
                 raise RuntimeError("Failed to install dependencies.")
+
+            # AiiDA plugins require reentry run to be found by AiiDA.
+            process = run_reentry_scan()
+            for line in io.TextIOWrapper(process.stdout, encoding="utf-8"):
+                stdout.write(line)
 
             # Restarting the AiiDA daemon to import newly installed plugins.
             process = run_verdi_daemon_restart()
--- a/aiidalab/app.py
+++ b/aiidalab/app.py
@@ -344,10 +344,10 @@ class _AiidaLabApp:
             # In practice this should never happen
             if installed_aiida is None:
                 return False
-            aiida1 = Requirement("aiida-core<2.0.0b1")
+            aiida_v1 = Requirement("aiida-core<2.0.0b1")
             # This is needed to handle pre-release versions such as 1.0.0b0
-            aiida1.specifier.prereleases = True
-            if installed_aiida.fulfills(aiida1):
+            aiida_v1.specifier.prereleases = True
+            if installed_aiida.fulfills(aiida_v1):
                 return True
             return False
 
--- a/aiidalab/app.py
+++ b/aiidalab/app.py
@@ -356,6 +356,7 @@ class _AiidaLabApp:
             # https://www.endpoint.com/blog/2015/01/getting-realtime-output-using-python/
 
             # Install package dependencies.
+            stdout.write(f"Running 'pip install --user {' '.join(args)}'\n")
             process = run_pip_install(*args, python_bin=python_bin)
             for line in io.TextIOWrapper(process.stdout, encoding="utf-8"):
                 stdout.write(line)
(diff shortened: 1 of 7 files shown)


aiidalab/utils.py
```diff
@@ -243,6 +243,14 @@ def run_pip_install(*args, python_bin=sys.executable):  # type: ignore
     )
 
 
+def run_reentry_scan():  # type: ignore
+    return subprocess.Popen(
```
unkcpz, line 247: If we want to make sure this can be run, we need `reentry` installed explicitly https://pypi.org/project/reentry/, correct? So it should be a dependency?

unkcpz: Moreover, instead of calling it with subprocess, I think it makes more sense to call it with API directly:

```python
from reentry import manager
manager.scan()
```

Not 100% sure this is exactly the `reentry scan`, you need have a double-check.

aiidalab/app.py
```diff
@@ -333,21 +334,47 @@ def find_dependencies_to_install(
...
         """Try to install the app dependencies with pip (if specified)."""
 
+        def _should_run_reentry():
+            # We do not need to run reentry scan for aiida-core>=2"""
+            packages = find_installed_packages(python_bin=None)
+            installed_aiida = get_package_by_name(packages, "aiida-core")
+            # In practice this should never happen
+            if installed_aiida is None:
+                return False
+            aiida1 = Requirement("aiida-core<2.0.0b1")
```
unkcpz, line 347: I prefer `aiida_v1` as variable ;)

aiidalab/app.py
```diff
@@ -333,21 +334,47 @@ def find_dependencies_to_install(
...
+    def _install_dependencies(self, python_bin, stdout):
         """Try to install the app dependencies with pip (if specified)."""
 
+        def _should_run_reentry():
+            # We do not need to run reentry scan for aiida-core>=2"""
+            packages = find_installed_packages(python_bin=None)
+            installed_aiida = get_package_by_name(packages, "aiida-core")
+            # In practice this should never happen
+            if installed_aiida is None:
+                return False
```
unkcpz, line 346: In principle, this will never happen but in practice, this is what exactly happened for some leftover versions that didn't specify the version for aiida-core, e.g `aiidalab-widgets-base==1.0.0a1`. Since there is no side effect of running `reentry scan`, can we return True to always running this if not specified?

unkcpz (changes requested):
Thanks @danielhollas, some questions especially worth to try if it is possible to call reentry scan from API.

> Dropping dependency on aiida-core in setup.cfg.

Fully agree, I assume initially add this since not running reentry scan is not compatible with old plugins and apps, which is what this PR tries to address.

aiidalab/utils.py
danielhollas: > If we want to make sure this can be run, we need reentry installed explicitly https://pypi.org/project/reentry/, correct? So it should be a dependency?

Well, you are correct, but we do not want to add that as a dependency, because it is not needed in the new Docker stack. So we implicitly rely on reentry being installed in the old Docker stack. Not sure what else we can do here.

unkcpz: Ah, I see your point, this makes sense. I think we shouldn't put complex in this backport support. I'll give it a test and have it.

aiidalab/app.py
danielhollas: @unkcpz this code path would only happen, if `aiida-core` is not installed. So this doesn't really depend on whether apps depend on `aiida-core` or not. TBH: if `aiida-core` is missing, we will fail anyway later when we try to run restart the daemon...

aiidalab/app.py
danielhollas: Done, thanks.

aiidalab/app.py
unkcpz: I see, I misunderstood the logic, it was checking with the packages not the requirements of the app. It looks good to me.

unkcpz (commented):
Thanks @danielhollas, the code looks good. I'll give it a test later today but I am not going to add unit test to implementation.

unkcpz:
I did the test, all works as expected except for the issue you mentioned that the update button is activated. I think we can get this one merged and test https://github.com/aiidalab/aiidalab/pull/368 based on this.

Another concern is about the `aiidalab/aiidalab-docker-stack`, it was pinning the `aiidalab` version where the `22.7.1` for the latest docker stack tag. Do we still pin the version or do we simply remove the constraint?

unkcpz (approved):

danielhollas:
> Another concern is about the aiidalab/aiidalab-docker-stack, it was pinning the aiidalab version where the 22.7.1 for the latest docker stack tag. Do we still pin the version or do we simply remove the constraint?

I think we should still pin the version, and just update it when we release a new one and publish a new `aiidalab/aiidalab-docker-stack:latest`. I opened a separate issue for that here: https://github.com/aiidalab/aiidalab-docker-stack/issues/363

PR #356, opened 2023-02-24
feat: mypy typing check for app.py and environment.py and utils.py

### LINKED ISSUE #354: Add type annotations and mypy as pre-commit hook

unkcpz:
fixes #354 

Simply did the typing check by mypy. If there are errors that are not related to requirements and packages, I didn't manage to fix those.


danielhollas:
@unkcpz thanks for this. What is your plan here, do you want to merge this after or before #347?

In any case it would be great to merge this before we release a new version, to have some confidence that there are no silly bugs. It migh be nice to do a similar subsequent PR on aiidalab-home.

danielhollas:
I would do the fixes in a separate PR, otherwise it's going to be a mess.

unkcpz:
@danielhollas I think it is ready to be reviewed. I give up on rebase from main but simply merge it. This leaves many commits. You can review directly as one commit.

aiidalab/app.py
```diff
@@ -523,7 +530,7 @@ class AiidaLabAppWatch:
             The AiidaLab app to monitor.
     """
 
-    class AppPathFileSystemEventHandler(FileSystemEventHandler):
+    class AppPathFileSystemEventHandler(FileSystemEventHandler):  # type: ignore
```
danielhollas, line 533: Why is this ignore needed?

.pre-commit-config.yaml
```diff
@@ -55,3 +50,19 @@ repos:
...
+    - repo: https://github.com/pre-commit/mirrors-mypy
+      rev: v0.991
+      hooks:
+          - id: mypy
+            args: [--config-file=setup.cfg]
+            additional_dependencies:
+                - types-cachetools
+                - types-requests
+                - types-cachetools
+                - types-toml
```
danielhollas, line 63: duplicate
```suggestion
                - types-cachetools
                - types-requests
                - types-toml
```

aiidalab/app.py
```diff
@@ -498,8 +505,8 @@ def install(
                     # Attempt rollback to previous version.
                     logger.info("Performing rollback to previously installed version.")
                     self._restore_from(trash_path)
-            except RuntimeError as error:
-                logger.warning(f"Rollback failed due to error: {error}!")
+            except RuntimeError as rerror:
```
danielhollas, line 508: This is a rather confusing naming. Why was this change needed?

danielhollas (approved):
@unkcpz This is great, thanks! I am not a mypy expert and didn't look too closely, but I like that you opened #363 so we can clean it up later. 

(only one minor problem I caught in the pre-commit configuration, you can merge after the fix)

aiidalab/app.py
unkcpz: Because there is another `error` lines ahead for another `RuntimeError`, the mypy complains this is not a good practice to override the local variable. I think using `type: igonre` here and replace it with `raise from` may be the correct solution.

aiidalab/app.py
unkcpz: It complains about "error: Class cannot subclass "FileSystemEventHandler" (has type "Any")  [misc]", I simply suppress it for the moment.

unkcpz (commented):
thanks @danielhollas! 
As I mentioned, we can get this one merged and fix more typing issues after. 
I am also not a mypy expert, this is the first time I bring mypy from scratch to a repository. Let's do it to #363 :wink:

aiidalab/app.py
unkcpz: Please see the latest commit, I change the name to `inner_error`.
The issue is that the `error` outside is override by the inner `RuntimeError` exception and in the `finally` clause the it will not report the correct exception if the inner RuntimeError raised.

danielhollas: @unkcpz thanks! I did not know about `raise ... from ...` syntax. I think that is a great solution in this case.
Go ahead with merging this.

I did notice one thing when looking at this code. The installation try block starting on line 465 is very big, and we're only catching the `RuntimeError` exceptions. Are we sure that there are no other failure modes coming from the functions that are called?
 
This is also most likely the direct reason for #334. When installing from command line, interrupting with Ctrl+C will interrupt the install without the rollback. I think until a more robust solution is implemented, as suggested by Sasha in #349, we should be catching `BaseException` instead of `RuntimeException`. Could you maybe take a look in another PR?

danielhollas:
Just copying a link to my comment so it does not get lost in the future.

https://github.com/aiidalab/aiidalab/pull/356#discussion_r1126309310

aiidalab/app.py
unkcpz: Yes, those seem related, I'll give it a test and see if the exception caught here can fix the issue #334

unkcpz:
fixes https://github.com/aiidalab/aiidalab/issues/354

Introduce the typing check by mypy. If there are errors that are not related to requirements and packages, I didn't manage to fix those in this commit but leave it for future fixes.

PR #357, opened 2023-02-25
add count to proper set busy traitlets

unkcpz:
This is the more direct cause of https://github.com/aiidalab/aiidalab-home/issues/100, rather than https://github.com/aiidalab/aiidalab/pull/349.
In the package, there is already a context manager `_show_busy` provide to hold the display before the operations are running. However, this context manager only set the `self.busy` to a bool, which will be overridden unexpectedly if the `_show_busy` is embedded in itself. This happened for `install_app`, `updata_app` and `uninstall_app`, they are supposed to use `_show_busy` to prevent information to be displayed, but inside these function also call `refresh` which have another `_show_busy`. I add a counter to properly set the `busy` traitlets. It is `False` (which means not busy) when it counts down to zero.

### DIFF
--- a/aiidalab/app.py
+++ b/aiidalab/app.py
@@ -8,6 +8,7 @@ import shutil
 import sys
 import tarfile
 import tempfile
+import threading
 from contextlib import contextmanager
 from dataclasses import asdict, dataclass, field
 from enum import Enum, Flag, auto
@@ -606,6 +607,7 @@ class AiidaLabApp(traitlets.HasTraits):
         self.name = self._app.name
 
         self._busy_count = 0
+        self._busy_count_lock = threading.Lock()
 
         try:
             self.logo = self._app.metadata["logo"]
@@ -654,11 +656,18 @@ class AiidaLabApp(traitlets.HasTraits):
     def _show_busy(self):
         """Apply this decorator to indicate that the app is busy during execution."""
         self.set_trait("busy", True)
-        self._busy_count += 1
+
+        # we need to use a lock here, because the busy trait is not thread-safe
+        # we may use _show_busy in different threads, e.g. when installing and auto-status refresh
+        with self._busy_count_lock:
+            self._busy_count += 1
+
         try:
             yield
         finally:
-            self._busy_count -= 1
+            with self._busy_count_lock:
+                self._busy_count -= 1
+
             if self._busy_count == 0:
                 self.set_trait("busy", False)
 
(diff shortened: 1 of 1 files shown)


danielhollas:
@unkcpz Thanks, this is a great catch! 

The code looks okay, but I think we need a Lock to guard access to the `self._busy_count` to ensure it is accessed atomically in case there are more threads.

https://stackoverflow.com/a/40072999/3682277

unkcpz:
The thread lock was added and the test was all good. You can easily test it by comparing branches with installing QeApp version `22.08.0` without a needed dependency installed e.g by uninstall the `filelock`.

aiidalab/app.py
```diff
@@ -652,10 +656,20 @@ def _default_busy(self):  # pylint: disable=no-self-use
     def _show_busy(self):
         """Apply this decorator to indicate that the app is busy during execution."""
         self.set_trait("busy", True)
+
+        # we need to use a lock here, because the busy trait is not thread-safe
+        # we may use _show_busy in different threads, e.g. when installing and auto-status refresh
+        with self._busy_count_lock:
+            self._busy_count += 1
```
danielhollas, line 663:
```suggestion
        # we need to use a lock here, because the busy trait is not thread-safe
        # we may use _show_busy in different threads, e.g. when installing and auto-status refresh
        with self._busy_count_lock:
            self._busy_count += 1
            self.set_trait("busy", True)
```

aiidalab/app.py
```diff
@@ -652,10 +656,20 @@ def _default_busy(self):  # pylint: disable=no-self-use
...
+
         try:
             yield
         finally:
-            self.set_trait("busy", False)
+            with self._busy_count_lock:
+                self._busy_count -= 1
+
+            if self._busy_count == 0:
+                self.set_trait("busy", False)
```
danielhollas, line 672:
```suggestion
            with self._busy_count_lock:
                self._busy_count -= 1
                if self._busy_count == 0:
                    self.set_trait("busy", False)
```

danielhollas (changes requested):
@unkcpz Thanks. I think however that the `set_trait("busy",x)` calls need to be behind the lock as well. Otherwise one thread can come in, set the trait to True at the top, while the other thread will subsequently set it to False at the bottom.

(PS: Thanks for letting me know how to test this, will do that once these changes are made)

aiidalab/app.py
danielhollas: Thanks. If it is within the lock I think the order does not matter. I will test this PR now.

danielhollas (approved):
Thanks @unkcpz, I did the test you suggested above and it worked!

PR #359, opened 2023-03-01
Increate refresh interval

### LINKED ISSUE #358: The app manager keeps on refreshing because of watchdog spamming opened event check
Once the app is installed, the it become "attached" to the app manager and the app manager will keep on refresh to get the latest version information of the app, e.g if the app is latest or if the dependencies are still compatible.. This status refresh is handled by the `AiidaLabApp.refresh` which is running in an independent thread in the background so it is not blocking other operations. 
However, as shown in the recording below, the loading spinner will show every 1 second and when it refreshes all other widgets are blocked and not interactive. 

EDIT: I forget to attach the screencast, thank you for the noticing @danielhollas 
Peek 2023-03-01 17-43

https://github.com/aiidalab/aiidalab/blob/798b459990a011308c7e14e01d0e295d05345142/aiidalab/app.py#L768-L773

unkcpz:
fixes #358 

Increase the time of refresh interval. However, it is that simple by just increase the number. Because in the `throttled` decorator, it will check the function it decorated and if the refresh is in running, it will not run again. This bring the issue that when the refresh_async is running in a saparated thread, the regular refresh triggered by the user operation such as changing the version to installed or pre-releases checkbox tick is not run since it is the same function which blocked in throttled decorator. 

Unfortunately, the changes in this PR still not perfectly working. I need check more how to make it work properly.


unkcpz:
Okay, this one is ready I think. But in case we slow down the https://github.com/aiidalab/aiidalab/pull/347, I suggest to start review this one after https://github.com/aiidalab/aiidalab/pull/347.

yakutovicha (approved):

danielhollas:
@unkcpz you mentioned in the other PR there were some issues with the tests. Was that resolved here? In any case let's merge #347 first and then rebase this.

danielhollas (commented):
I've been looking at the code and how the `async_refresh` is used and I am not fully convinced that the solution in this PR is the correct one. I'll need to take a deeper look.

At the very least, removing the `throttled` decorator from the `refresh` method might have unintended consequences since this method is called directly from a bunch of places...

unkcpz:
The problem comes from `watchdog` where we use it in `AiidalabAppWatch`, I close this one and open other PR for fix.

PR #362, opened 2023-03-03
watchdog file system check not trigger refresh on file opened

### LINKED ISSUE #358: The app manager keeps on refreshing because of watchdog spamming opened event check

unkcpz:
fixes #358 

The watchdog file system monitor introduce a new event handler `INOPNED` for when file opened. The new "file opened" event is included by default in all events: https://github.com/gorakhargosh/watchdog/blob/v2.3.0/src/watchdog/events.py#L289

The details are discussed at https://github.com/gorakhargosh/watchdog/issues/949, we encounter the same issue. But we use `on_any_event` explicitly, we adopt the same change as https://github.com/gorakhargosh/watchdog/commit/25a2d1fec55fc32f97055ba827c8ce7b959acb34 to avoid spammed by the opened event update.

### DIFF
--- a/aiidalab/app.py
+++ b/aiidalab/app.py
@@ -27,7 +27,7 @@ import traitlets
 from dulwich.errors import NotGitRepository
 from packaging.requirements import Requirement
 from packaging.version import parse
-from watchdog.events import FileSystemEventHandler
+from watchdog.events import EVENT_TYPE_OPENED, FileSystemEventHandler
 from watchdog.observers import Observer
 from watchdog.observers.polling import PollingObserver
 
@@ -530,8 +530,9 @@ class AiidaLabAppWatch:
             self.app = app
 
         def on_any_event(self, event):
-            """Refresh app for any event."""
-            self.app.refresh_async()
+            """Refresh app for any event except opened."""
+            if event.event_type != EVENT_TYPE_OPENED:
+                self.app.refresh_async()
 
     def __init__(self, app):
         self.app = app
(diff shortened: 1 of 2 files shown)


danielhollas:
Oh, this is a great catch, thanks! Since this is a new event,  could you please check in which version was it added, and bump the minimum watchdog requirements in `setup.cfg` accordingly.

danielhollas:
Thanks for chiming in @BoboTiG : 💖 

Do I understand correctly that the 'breaking' change was reverted in 2.3.1? Even if that is the case, I think the change in this PR still makes sense for us. 🤔

BoboTiG:
That PR still makes sense 👍🏻 
I reverted the behavior change for `watchmedo` only.

danielhollas (approved):

PR #367, opened 2023-03-09
Using event to log the stop status of the monitor threading

### LINKED ISSUE #361: Fix invalid `stop_flag` attribute.
The API of `Thread` class is change, there is no `stop_flag` attribute. This code is problematic but better to be taken care in other PR.

_Originally posted by @unkcpz in https://github.com/aiidalab/aiidalab/pull/356#discussion_r1117796839_

unkcpz:
fixes #361 

I did the minimal change by replace `_stop_flag` with the `threading.Event` object.

### DIFF
--- a/aiidalab/app.py
+++ b/aiidalab/app.py
@@ -660,14 +660,15 @@ class AiidaLabAppWatch:
         return self._monitor_thread and self._monitor_thread.is_alive()
 
     def join(self, timeout=None):
-        """Join the watch after stopping.
+        """Join the watch and observer after stopping.
 
         This function will timeout if a timeout argument is provided. Use the
         is_alive() function to determien whether the watch was stopped within
         the given timout.
         """
-        if self._monitor_thread is not None:
+        if self._monitor_thread is not None and self._observer is not None:
             self._monitor_thread.join(timeout=timeout)
+            self._observer.join(timeout=timeout)
 
 
 class AiidaLabApp(traitlets.HasTraits):  # type: ignore
--- a/aiidalab/app.py
+++ b/aiidalab/app.py
@@ -540,7 +540,6 @@ class AiidaLabAppWatch:
         def on_any_event(self, event):
             """Refresh app for any event except opened."""
             if event.event_type != EVENT_TYPE_OPENED:
-                print(event)
                 self.app.refresh_async()
 
     def __init__(self, app):
@@ -599,11 +598,8 @@ class AiidaLabAppWatch:
 
             def check_path_exists_changed():
                 is_dir = os.path.isdir(self.app.path)
-                print("or this??", is_dir)
-                print("checking...")
                 while not self._monitor_thread_stop.is_set():
                     switched = is_dir != os.path.isdir(self.app.path)
-                    print("??", switched)
                     if switched:
                         # this is for when the app folder first time create or deleted
                         is_dir = not is_dir
--- a/aiidalab/app.py
+++ b/aiidalab/app.py
@@ -540,6 +540,7 @@ class AiidaLabAppWatch:
         def on_any_event(self, event):
             """Refresh app for any event except opened."""
             if event.event_type != EVENT_TYPE_OPENED:
+                print(event)
                 self.app.refresh_async()
 
     def __init__(self, app):
@@ -598,9 +599,13 @@ class AiidaLabAppWatch:
 
             def check_path_exists_changed():
                 is_dir = os.path.isdir(self.app.path)
+                print("or this??", is_dir)
+                print("checking...")
                 while not self._monitor_thread_stop.is_set():
                     switched = is_dir != os.path.isdir(self.app.path)
+                    print("??", switched)
                     if switched:
+                        # this is for when the app folder first time create or deleted
                         is_dir = not is_dir
                         self.app.refresh()
 
--- a/aiidalab/app.py
+++ b/aiidalab/app.py
@@ -666,8 +666,9 @@ class AiidaLabAppWatch:
         is_alive() function to determien whether the watch was stopped within
         the given timout.
         """
-        if self._monitor_thread is not None and self._observer is not None:
+        if self._monitor_thread is not None:
             self._monitor_thread.join(timeout=timeout)
+        if self._observer is not None:
             self._observer.join(timeout=timeout)
 
 
--- a/aiidalab/app.py
+++ b/aiidalab/app.py
@@ -524,6 +524,8 @@ class AiidaLabAppWatch:
     AiiDAlab app. This is achieved by monitoring the app repository
     for existance and changes.
 
+    If there is a change in the app repository, the app is refreshed.
+
     Arguments:
         app (AiidaLabApp):
             The AiidaLab app to monitor.
@@ -546,6 +548,7 @@ class AiidaLabAppWatch:
         self._started = False
         self._monitor_thread = None
         self._observer = None
+        self._monitor_thread_stop = threading.Event()
 
     def __repr__(self):
         return f"<{type(self).__name__}(app={self.app!r})>"
@@ -595,7 +598,7 @@ class AiidaLabAppWatch:
 
             def check_path_exists_changed():
                 is_dir = os.path.isdir(self.app.path)
-                while not self._monitor_thread.stop_flag:  # type: ignore
+                while not self._monitor_thread_stop.is_set():
                     switched = is_dir != os.path.isdir(self.app.path)
                     if switched:
                         is_dir = not is_dir
@@ -614,7 +617,7 @@ class AiidaLabAppWatch:
                     self._observer.stop()
 
             self._monitor_thread = Thread(target=check_path_exists_changed)
-            self._monitor_thread.stop_flag = False  # type: ignore
+            self._monitor_thread_stop.clear()
             self._monitor_thread.start()
 
         self._started = True
@@ -622,7 +625,7 @@ class AiidaLabAppWatch:
     def stop(self):
         """Stop watching the app repository for file system events."""
         if self._monitor_thread is not None:
-            self._monitor_thread.stop_flag = True
+            self._monitor_thread_stop.set()
 
     def is_alive(self):
         """Return True if this watch is still alive."""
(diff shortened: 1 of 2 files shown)


unkcpz:
To be honest, I have no idea have to test the code using threading. I tried something like:

```python
def test_app_watch(tmp_path):
    """Test the aiidalab app watch responsive to the app path changes."""
    
    @dataclass
    class TestApp:
        path: Path
        x: int = 0
        
        def refresh_async(self):
            print("refresh")
            self.x += 1
        
    app = TestApp(path=Path(tmp_path))
    app_watch = AiidaLabAppWatch(app)
    app_watch.start()
    
    # print(app_watch.app.x)
    
    # Trigger the app refresh by changing the app path
    d = tmp_path / "test"
    d.mkdir()
    
    # print(app_watch.app.x)
    app_watch.stop()
    app_watch.join()
    # print(app_watch.app.x)
    # print(app.x)
    # # app.refresh_async()
    # print(app.x)
```

But the behaviour is not very clear. In my understanding, the pytest run on on thread and it should block if there are more than one threading?

unkcpz:
This is how I test the change, I add a print clause at `on_any_event` before the `refresh_async`. I did the change in the app folder by creating or deleting a file. The print is executed and the text print on the screen.

danielhollas:
> But the behavior is not very clear. In my understanding, the pytest run on on thread and it should block if there are more than one threading?

I am not sure I understand, does pytest print any errors when running this code? Or does the code not work as expected?

In any case, we also need to test (manually or pytest) that the stop method works and actually stops the watcher thread.
(btw: I think when I looked last time, the stop method doesn't seem to be called anywhere, which is a little suspicious).

tests/test_appclass.py
```diff
@@ -39,3 +43,44 @@ def test_dependencies(generate_app):
...
+            print("refresh")
+            self.y += 1
+
+    app = DummyApp(path=Path(tmp_path))
+    app_watch = AiidaLabAppWatch(app)
+    app_watch.start()
+
+    # The observer is start in a thread so need to wait until it is alive
+    while app_watch._observer is None or not app_watch._observer.is_alive():
+        sleep(0.1)
```
unkcpz, line 70: This is the trick, I didn't do this so I cannot get the persistent result previously. The observer is launched in a thread, if it is not started, the file change will not trigger the action.

aiidalab/app.py
```diff
@@ -538,6 +540,7 @@ def __init__(self, app):
         def on_any_event(self, event):
             """Refresh app for any event except opened."""
             if event.event_type != EVENT_TYPE_OPENED:
+                print(event)
```
unkcpz, line 543:
```suggestion
```

aiidalab/app.py
```diff
@@ -595,9 +599,13 @@ def start(self):
 
             def check_path_exists_changed():
                 is_dir = os.path.isdir(self.app.path)
-                while not self._monitor_thread.stop_flag:  # type: ignore
+                print("or this??", is_dir)
+                print("checking...")
```
unkcpz, line 603:
```suggestion
```

aiidalab/app.py
```diff
@@ -595,9 +599,13 @@ def start(self):
...
+                while not self._monitor_thread_stop.is_set():
                     switched = is_dir != os.path.isdir(self.app.path)
+                    print("??", switched)
```
unkcpz, line 606:
```suggestion
```

tests/test_appclass.py
```diff
@@ -39,3 +43,44 @@ def test_dependencies(generate_app):
...
+
+
+def test_app_watch(tmp_path):
+    """Test the aiidalab app watch responsive to the app path changes."""
+
+    @dataclass
+    class DummyApp:
+        path: Path
+        x: int = 0
+        y: int = 0
```
danielhollas, line 55: Unused variable
```suggestion
```

Also the method `refresh` is not used.

tests/test_appclass.py
```diff
@@ -39,3 +43,44 @@ def test_dependencies(generate_app):
...
+    # touch a file will trigger action 4 times: create and close file and two modifies of folder
+    testfile = tmp_path / "test0"
+    testfile.touch()
+
+    app_watch.stop()
+    app_watch.join()
+
+    assert app.x == 4
+
+    # After the thread is stop and join the obsever is inactivate and will not trigger refresh_async
```
danielhollas, line 82:
```suggestion
    # After the observer is stopped, file system events should no longer trigger `refresh_async`
```

tests/test_appclass.py
```diff
@@ -39,3 +43,44 @@ def test_dependencies(generate_app):
...
+    testfile = tmp_path / "test1"
+    testfile.touch()
+
```
danielhollas, line 85: I wonder if there should be a sleep statement before this assert to really test this. 
Otherwise there could be a race condition, and the thread is still running

tests/test_appclass.py
```diff
@@ -39,3 +43,44 @@ def test_dependencies(generate_app):
...
+    while app_watch._observer is None or not app_watch._observer.is_alive():
+        sleep(0.1)
+
+    # Trigger action by file events
+    # touch a file will trigger action 4 times: create and close file and two modifies of folder
+    testfile = tmp_path / "test0"
+    testfile.touch()
+
+    app_watch.stop()
+    app_watch.join()
```
danielhollas, line 78: Should we somehow verify that the thread was stopped before proceeding? Maybe a while loop with `is_alive` as in the code above? Otherwise we risk creating a flaky test.

danielhollas (changes requested):
@unkcpz awesome! Thanks so much for figuring out a test for this, that's really great. 
(by the way, would this new test fail before the code change in this PR?) 

The code itself looks good, I have a couple comments on the test. 
I haven't tested this manually yet, will do tomorrow or Monday.

tests/test_appclass.py
unkcpz: Yes, I remove `y` but I think it is better to keep `refresh` method. It is not hinted by this test but it is inside the thread target function `check_path_exists_changed`.

unkcpz:
> btw: I think when I looked last time, the stop method doesn't seem to be called anywhere, which is a little suspicious

From my understanding, the `stop` and `join` are not called in the AiiDAlab, since the thread should keep on running until the main thread the web page is closed. There is no need to stop that monitor thread manually in practical use case. 
But it is useful for test, since if I don't call `stop` and `join` in the unit test, the launched side thread will keep on running and the test is stuck. 

> by the way, would this new test fail before the code change in this PR?

No, the test still pass. The initial implementation with `_stop_flag` is not wrong, the `Thread` object can set the value on the fly. Only the mypy will complain. There is other way such as checking `self._monitor_thread_stop` thread is not `None` etc. But I think using `Event` here is a proper solution.

tests/test_appclass.py
unkcpz: I don't think this is needed, the `app_watch.join()` will make sure all the thread is joined, no?

tests/test_appclass.py
unkcpz: Okay, I see your point, I did this by joining the `_observer` thread after stopping. Please check the latest commit.

aiidalab/app.py
```diff
@@ -641,29 +645,30 @@ def check_path_exists_changed():
...
+        """Join the watch and observer after stopping.
 
         This function will timeout if a timeout argument is provided. Use the
         is_alive() function to determien whether the watch was stopped within
         the given timout.
         """
-        if self._monitor_thread is not None:
+        if self._monitor_thread is not None and self._observer is not None:
             self._monitor_thread.join(timeout=timeout)
+            self._observer.join(timeout=timeout)
```
danielhollas, line 671: I think it is safer to split this condition, otherwise you risk that the monitor thread started but observer thread is not started, and the join statements will not execute.

```suggestion
        if self._monitor_thread is not None:
            self._monitor_thread.join(timeout=timeout)
        if self._observer is not None:
            self._observer.join(timeout=timeout)
```

tests/test_appclass.py
```diff
@@ -35,3 +39,42 @@ def test_dependencies(generate_app):
...
+    app_watch = AiidaLabAppWatch(app)
+    app_watch.start()
+
+    # The observer is start in a thread so need to wait until it is alive
+    while app_watch._observer is None or not app_watch._observer.is_alive():
+        sleep(0.1)
+
+    # Trigger action by file events
+    # touch a file will trigger action 4 times: create and close file and two modifies of folder
+    testfile = tmp_path / "test0"
```
danielhollas, line 65: Wow, I did not know the Path object supports this syntax, this is pretty cool!

tests/test_appclass.py
```diff
@@ -35,3 +39,42 @@ def test_dependencies(generate_app):
...
+    testfile.touch()
+
+    app_watch.stop()
+    app_watch.join()
```
danielhollas, line 69: Thanks. I think it is safer to use a timeout and then assert that the two thread stopped, otherwise we risk a potential deadlock in our CI. We can also check the number of threads before and after. 

```suggestion
    assert threading.active_count() > 1

    app_watch.stop()
    app_watch.join(timeout=10.)
    assert threading.active_count() == 1
```

danielhollas (changes requested):
Thanks @unkcpz! I've tested this and it seems to work.
I just have two small ask but otherwise looks good.

tests/test_appclass.py
unkcpz: Nice advice, more robust for the test involved with threading. changed.

unkcpz:
> Huh, today I learned that you can attach arbitrary attributes to an object. exploding_head I was so confused how this code ever worked, now I get it.

I am from c/c++ community so never use the very pythonic way before. I think the mypy with type hint more or less move python to the strong typed language which I think is very good.

tests/test_appclass.py
danielhollas: Did you push the change? I don't see it on Github.

unkcpz: Gosh, what a failure. I push now. 😢

danielhollas (approved):
Great, let's go! :rocket:

PR #369, opened 2023-03-12
CLI: Add -h option as an alias for --help

danielhollas:
Based on user feedback at the last AiiDA hackathon,
`verdi` command now supports `-h` option as an alias for `--help`, as of https://github.com/aiidateam/aiida-core/pull/5792

Let's do the same thing here.

If you agree with this, I'll submit the same PR for `aiidalab-launch` as well.

### DIFF
--- a/aiidalab/__main__.py
+++ b/aiidalab/__main__.py
@@ -68,7 +68,7 @@ def _list_apps(apps_path):
         )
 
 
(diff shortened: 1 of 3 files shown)


unkcpz (approved):
Thanks @danielhollas, LGTM.

Issue #364, opened 2023-03-06
Make installation workflow more robust


danielhollas:
The brittleness of the installation workflow partly comes from the fact that the app is installed before the dependencies. That is not common. Pip and conda do exactly the opposite. They first install the dependencies, and only then proceed with installing the main package.

I am thinking that reverting the order might solve a lot of things for us too.

We need to pull the folder to a temporary location (already implemented) and install dependencies(already implemented). If that was successful - copy the folder to `~/apps`. It is really about reshuffling the code a tiny bit.

_Originally posted by @yakutovicha in https://github.com/aiidalab/aiidalab/issues/349#issuecomment-1425640565_

danielhollas:
Just to illustrate the importance of this, I was able to trigger the installation of an App twice by clicking the Install button twice in rapid succession.

PR #370, opened 2023-03-14
Catch all exceptions when installing app

### LINKED ISSUE #334: aiidalab does not detect interrupted app installation

danielhollas:
### Summary

Previously, we were only catching RuntimeError which left a lot of room for other failures during installation, which would result in an incomplete installation of an app. Generally, there's no way of knowing after-the-fact that the installation ended prematurely, e.g. `aiidalab list` would show the app as installed.

Note that this includes for example `post_install` script failures.

Now we catch everything, including Keyboard interrupts, so that we always trigger the already existing rollback mechanism.

See issue #334 for details of how to reproduce this.

### Test plan

1. Trigger installation of an app from CLI `aiidalab -v install aurora==0.2.1`
2. Interrupt the installation in the middle with Ctrl+C
3. The installation should rollback. Verify with `aiidalab list` that the app is not installed.
4. Try installing an app from a following URL, where I specifically modified the `post_install` script to fail. `aiidalab -v install aiidalab-ispg@git+https://github.com/ispg-group/aiidalab-ispg@post-install-fail`
5. Failing post_install should now trigger rollback as well.

Closes #334

Note that the more robust fix to this should be done in #364

### DIFF
--- a/aiidalab/app.py
+++ b/aiidalab/app.py
@@ -356,7 +356,7 @@ class _AiidaLabApp:
             # https://www.endpoint.com/blog/2015/01/getting-realtime-output-using-python/
 
             # Install package dependencies.
-            stdout.write(f"Running 'pip install --user {' '.join(args)}'\n")
+            logger.info(f"Running 'pip install --user {' '.join(args)}'\n")
             process = run_pip_install(*args, python_bin=python_bin)
             for line in io.TextIOWrapper(process.stdout, encoding="utf-8"):
                 stdout.write(line)
@@ -366,7 +366,7 @@ class _AiidaLabApp:
 
             # AiiDA plugins require reentry run to be found by AiiDA.
             if _should_run_reentry():
-                stdout.write("\nRunning 'reentry scan'\n")
+                logger.info("\nRunning 'reentry scan'\n")
                 process = run_reentry_scan()
                 for line in io.TextIOWrapper(process.stdout, encoding="utf-8"):
                     stdout.write(line)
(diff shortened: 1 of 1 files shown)


unkcpz (commented):
@danielhollas thanks for the fix, the changes look good. 

Since we have more experience with adding tests for this class, it is worth trying to monkeypatch a bunch of stuff in the `install` method and hint the exception to check it "work" as expected. If it is too complex we can leave it to the future.

danielhollas:
Thanks for taking a look. I've looked into wrinting a test for the install workflow and it was a bit more work than I have time at the moment. I do not think it makes sense to test this functionality in isolation, we need to test the whole installation workflow comprehensively, and I do not have a bandwidth for that in the next three weeks unfortunately.

danielhollas:
@yakutovicha can you take a brief look? Just want to verify that you agree with the approach I took here.

unkcpz (approved):
>  I do not think it makes sense to test this functionality in isolation, we need to test the whole installation workflow comprehensively, and I do not have a bandwidth for that in the next three weeks unfortunately.

No problem. I approved it, see if @yakutovicha want to have another look.

yakutovicha (approved):
I agree with the change 👍 . Please go ahead with merging it.

PR #372, opened 2023-03-14
Add name of Package as the property

unkcpz:
This change is required by https://github.com/aiidalab/aiidalab-widgets-base/pull/446. Since we change the Package.name to `_name` in https://github.com/aiidalab/aiidalab/pull/347.


unkcpz:
As mentioned if what was discussed in https://github.com/aiidalab/aiidalab-widgets-base/pull/446#issuecomment-1469142763 is adopted, this change is not needed anymore. I close this one.

PR #374, opened 2023-03-16
incompatibility check for app detached will go through the local repo

### LINKED ISSUE #373: Detached app shows spurious warning
I noticed when I install AWB, and then modify the repository somehow, e.g. by switching to a different branch while developing, the app incompatible status is displayed, even though all dependencies are installed.

image

Not sure if the problem is here or in `aiidalab-home`. Note that this issue does not occur for an unregistered app.

Perhaps there needs to be an extra logic, where for detached apps we should parse dependencies from local `setup.cfg` and not from the registry?

unkcpz:
fixes #373 

When the app is installed but the app repo gets new commit, the app version will become UNKNOWN. Therefore, the requirements list cannot check with the version from the registered list. Rather the requirements list needs to access from the local repo. 
Check requirements upon local repo happened for the app not registered but also should be for the app detached. The condition is added for detached app for incompatibility check.

### DIFF
--- a/aiidalab/app.py
+++ b/aiidalab/app.py
@@ -287,11 +287,12 @@ class _AiidaLabApp:
         return self.remote_update_status() == AppRemoteUpdateStatus.DETACHED
 
     def find_incompatibilities(self, version, python_bin=None):
-        """Compatibility is checked by comparing between the requirement list of package
+        """Compatibility is checked by comparing the app requirements
         with the packages installed in the python environment.
-        If the app is registried which means the requirments list can get for certain version only,
-        it will checked w.r.t that. If the app is not registried or it is detached which usually caused from
-        the app is modified locally the requirements list will directly read from the local repository.
+
+        If the app is registered the list of requirements is fetched from the registry for the specific version.
+        If the app is not registered or if it is detached (i.e. locally modified),
+        the requirements list is read from the local repository (e.g. by parsing setup.cfg).
         """
         if python_bin is None:
             python_bin = sys.executable
@@ -863,11 +864,7 @@ class AiidaLabApp(traitlets.HasTraits):  # type: ignore
         return None
 
     def _is_compatible(self, app_version: str) -> bool:
-        """Determine whether the specified version is compatible.
-
-        Note: if the app_version is UNKNOWN, then the compatibility is indetermined and False is returned.
-        This usually happens when the app is detached.
-        """
+        """Determine whether the specified version is compatible."""
         try:
             incompatibilities = dict(
                 self._app.find_incompatibilities(version=app_version)
--- a/aiidalab/app.py
+++ b/aiidalab/app.py
@@ -283,12 +283,21 @@ class _AiidaLabApp:
             elif not pkg.fulfills(requirement):
                 yield requirement
 
+    def is_detached(self):
+        """Check whether the app is detached from the registry."""
+        return self.remote_update_status() == AppRemoteUpdateStatus.DETACHED
+
     def find_incompatibilities(self, version, python_bin=None):
+        """Compatibility is checked by comparing between the requirement list of package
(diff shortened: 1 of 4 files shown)


aiidalab/app.py
```diff
@@ -125,7 +125,6 @@ def from_id(cls, app_id, registry_entry=None, apps_path=None):
 
         return cls.from_registry_entry(path=app_path, registry_entry=registry_entry)
 
-    @property
```
danielhollas, line 128: Why did you change this?

To me, `is_registered` is an immutable property of the app (unlike `is_installed` or `is_detached` that may change due to user action) so it makes sense for it to be `@property`

aiidalab/app.py
```diff
@@ -283,12 +282,21 @@ def _find_incompatibilities_python(
...
+    def is_detached(self):
+        """Check whether the app is detached from the registry."""
+        return self.remote_update_status() == AppRemoteUpdateStatus.DETACHED
+
     def find_incompatibilities(self, version, python_bin=None):
+        """Compatibility is checked by comparing between the requirement list of package
+        with the packages installed in the python environment.
+        If the app is registried which means the requirments list can get for certain version only,
+        it will checked w.r.t that. If the app is not registried or it is detached which usually caused from
+        the app is modified locally the requirements list will directly read from the local repository.
```
danielhollas, line 294:
```suggestion
        """Compatibility is checked by comparing the app requirements
        with the packages installed in the python environment.
 
        If the app is registered the list of requirements is fetched from the registry for the specific version.
        If the app is not registered or if it is detached (i.e. locally modified),
        the requirements list is read from the local repository (e.g. by parsing setup.cfg).
```

aiidalab/app.py
```diff
@@ -855,7 +863,11 @@ def _default_compatible(self) -> None:  # pylint: disable=no-self-use
         return None
 
     def _is_compatible(self, app_version: str) -> bool:
-        """Determine whether the specified version is compatible."""
+        """Determine whether the specified version is compatible.
+
+        Note: if the app_version is UNKNOWN, then the compatibility is indetermined and False is returned.
+        This usually happens when the app is detached.
```
danielhollas, line 869: Leftover comment?

```suggestion
```

tests/test_appdataclass.py
```diff
@@ -147,3 +147,43 @@ def test_update_status_latest_version_incompatible(
...
+    # because the Environment.scan method is used for check the compatibility of local repo.
+
+    # monkeypatch the scan method to return a fake environment
+    # the fake environment has the aiida-core version that is compatible with the app
+    monkeypatch.setattr(
+        Environment,
+        "scan",
+        lambda _: Environment(python_requirements=["aiida-core~=2.0"]),
+    )
+    assert len(list(aiidalab_app_data.find_incompatibilities("v0.1.0"))) == 0
```
danielhollas, line 181: Could we test this on a higher level? Perhaps testing the `is_compatible` attribute?

danielhollas (changes requested):
@unkcpz thanks! Just some minor comments. I will give it a test soon when I get to work.

aiidalab/app.py
unkcpz: Yes, deleted.

aiidalab/app.py
danielhollas: Okay. I don't have a strong feeling about this. Just in general I would like to minimize unrelated changes in PRs, since it makes the review harder and because Python is not a compiled language there's always a risk of introducing bugs if one is not careful. It's best if refactoring like this were done separately.

danielhollas (approved):

aiidalab/app.py
unkcpz: > It's best if refactoring like this were done separately.

Okay, Will do it next time.

PR #376, opened 2023-03-17
Officially drop support for Python 3.7, test 3.11

danielhollas:
We've stopped testing Python 3.7 a while ago, and "dropped" the support in #305, but forgot to actually update `setup.cfg`.

Here's a list of new features in Python 3.8, most notably the walrus operator and the `f"{var=}"` debug syntax for  f-strings. I've verified with grep that we've not used these features yet.

https://docs.python.org/3/whatsnew/3.8.html#summary-release-highlights


unkcpz (approved):
All good. Thanks @danielhollas

Issue #371, opened 2023-03-14
Links in the base registry template are wrong


yakutovicha:
There are still wrong links in the currently used template, see here, for instance:
https://github.com/aiidalab/aiidalab/blob/6acf490983e63af3075b79af86cd698a7db27aa9/aiidalab/registry/templates/base.html#L3

## TODO:
- [ ] Identify all the wrong links
- [ ] fix them

danielhollas:
I think longer term it would make sense to integrate this page in the new website at `aiidalab.net`

danielhollas:
Also the link just below should be set to the new aiidalab website at `aiidalab.net`

PR #379, opened 2023-03-22
Fix base.html URLs

Mahhheshh:
Solves #371 
Updated the URLs used in the footer of the application to use the correct links. Previously, the links were pointing to the wrong resources, but have been updated to point to the correct URLs. This includes updating the links to the AiiDAlab registry page in the aiidalab-registry repository. These changes ensure that users can access the correct resources from the footer of the application.

This pull request fixes an issue in the parent file `base.html` at https://github.com/aiidalab/aiidalab/blob/50498b08ea77937cb8af0e8d753446dd17b6d637/aiidalab/registry/templates/base.html#L3 which was causing issues in the child file `layout.html` at https://github.com/aiidalab/aiidalab-registry/blob/8bc5d00fffa2016f562d065d81bd5a111c99479e/src/templates/layout.html#L8. The issue was causing variables to not be overwritten in the registry repository. The changes made in this pull request fix the links in the footer, and ensure that the variables can be correctly overwritten in the child file.

### DIFF
--- a/docs/source/usage/index.rst
+++ b/docs/source/usage/index.rst
@@ -9,7 +9,7 @@ Access AiiDAlab
 
 As a user, you have three options to access AiiDAlab:
 
- 1. Log into one of the `open AiiDAlab servers <https://materialscloud.org/aiidalab>`_.
+ 1. Log into one of the `open AiiDAlab servers <https://www.aiidalab.net/deployments/>`_.
  2. Run the AiiDAlab docker container directly :ref:`on your local machine <usage:run-locally>`.
  3. Download the `Quantum Mobile Virtual Machine <https://quantum-mobile.readthedocs.io/>`_, open a terminal and run ``aiidalab``.
 
--- a/aiidalab/registry/templates/base.html
+++ b/aiidalab/registry/templates/base.html
@@ -1,7 +1,7 @@
 <!doctype html>
 
-{% set app_registry_url = "http://aiidateam.github.io/aiida-registry" %}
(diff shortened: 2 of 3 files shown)


yakutovicha (changes requested):
thanks, @Mahhheshh. Could you also fix the links in: `docs/source/conf.py` and `docs/source/usage/index.rst`? They still point to the old AiiDAlab URL.

yakutovicha:
> Yes, are there any additional links I'm missing?

I think that is all.

docs/source/usage/index.rst
```diff
@@ -9,7 +9,7 @@ Access AiiDAlab
 
 As a user, you have three options to access AiiDAlab:
 
- 1. Log into one of the `open AiiDAlab servers <https://materialscloud.org/aiidalab>`_.
+ 1. Log into one of the `open AiiDAlab servers <https://www.aiidalab.net/>`_.
```
yakutovicha, line 12:
```suggestion
 1. Log into one of the `open AiiDAlab servers <https://www.aiidalab.net/deployments/>`_.
```

Pipfile
```diff
@@ -0,0 +1,11 @@
+[[source]]
```
yakutovicha, line 1: I don't think this file is needed.

yakutovicha (changes requested):
Thanks, just a few small changes.

Pipfile
Mahhheshh: Sorry for that I will remove it

yakutovicha (approved):

Issue #342, opened 2022-12-14
Support pyproject.toml installs in AiiDAlab apps


danielhollas:
Goal is for the AiiDAlab applications to be able to move from `setup.cfg` to `pyproject.toml` configuration.
I believe @unkcpz said that this currently does not work.

rough TODO list:

- [ ] Update docs https://github.com/aiidalab/aiidalab/blob/main/docs/source/app_development

- [ ] support parsing of `pyproject.toml` in `metadata.py` and `environment.py`

- [ ] fix the logic here, we shouldn't require file `setup.py` to be present https://github.com/aiidalab/aiidalab/blob/69b0738b113c75e772d5bb89597ad330966d0b81/aiidalab/app.py#L287

PR #382, opened 2023-05-02
Allow for removing setup.py in favour of simple pyproject.toml

danielhollas:
Step toward #342.

In this PR, we still require the `setup.cfg`, but instead of dummy `setup.py`, we can now user a rudimentary `pyproject.toml` to declare the build dependencies explicitly.

### DIFF
--- a/aiidalab/app.py
+++ b/aiidalab/app.py
@@ -395,7 +395,10 @@ class _AiidaLabApp:
         for path in (self.path.joinpath(".aiidalab"), self.path):
             if path.exists():
                 try:
-                    if path.joinpath("setup.py").is_file():
+                    if (
+                        path.joinpath("setup.py").is_file()
+                        or path.joinpath("pyproject.toml").is_file()
+                    ):
                         _pip_install(str(path), stdout=stdout)
                     elif path.joinpath("requirements.txt").is_file():
                         _pip_install(


danielhollas:
For testing this, you can run `aiidalab install aiidalab-ispg@git+https://github.com/ispg-group/aiidalab-ispg@remove-setup.py`, to see that the dependencies are still being installed even without `setup.py`

unkcpz:
thanks @danielhollas! The code here at https://github.com/aiidalab/aiidalab/blob/3972627b04cc81383e86c6517691e61dfb5074fe/aiidalab/environment.py#L55-L72 is relevant, and I think we can add the support by using [pyproject-api](https://pyproject-api.readthedocs.io/en/latest/). 

In terms of tests, the crucial part is the requirements information can be parsed from app url. I am keen to have unit test for it by mock the `setup.cfg` and `pyproject.toml`.

danielhollas:
> is relevant, and I think we can add the support _parse_pyproject by using [pyproject-api](https://pyproject-api.readthedocs.io/en/latest/).

For sure. But just to reiterate, the purpose of this PR is NOT a full support of `pyproject.toml`. We still have the app dependencies in `setup.cfg`. No parsing code is changed. Here I simply want to be able to get rid of the superfluous `setup.py` file.

PR #383, opened 2023-05-22
support using _ for matching all branches in getting registry

unkcpz:
In order to get all tags from different branches by setting in `aiidalab-registry` with:

```
releases:
  - url: 'git+https://github.com/aiidalab/aiidalab-qe.git@_:v22.05.0^..'
```

### DIFF
--- a/aiidalab/registry/releases.py
+++ b/aiidalab/registry/releases.py
@@ -29,16 +29,61 @@ def _split_release_line(url):
         return urlunsplit(parsed_url._replace(path=path)), release_line
     return url, None
 
+def _get_tags(repo, rev, rev_selection):
+    # While the git rev-list command supports listing revisions for a
+    # single ref, in this context we only support rev selections for a
+    # range, not for individual refs.
+    if ".." not in rev_selection:
+        raise ValueError(
+            "The rev_selection '{rev_selection}' must specify a range, "
+            "that means must contain the range operator '..'."
+        )
+
+    # Incomplete revision selections such as `@main:v1..` must be expanded to
+    # `@main:v1..main`. Therefore, we first determine the branch ref for the
+    # given rev:
+    for ref in [f"refs/heads/{rev}", f"refs/remotes/origin/{rev}"]:
+        if ref.encode() in repo.refs:
+            break
+    else:
+        raise RuntimeError(f"Revision '{rev}' not a valid branch name.")
+
+    # Transform a potentially incomplete rev_selection into one that
+    # contains the branch ref. For example `v1..` is expanded to
+    # `v1..{ref}`, where `{ref}` is replaced with the actual reference.
+    start, _, stop = rev_selection.rpartition("..")
+    selected_commits = repo.rev_list(f"{start or ref}..{stop or ref}")
+
+    for tag in repo.get_merged_tags(rev):
+        commit = repo.get_commit_for_tag(tag)
+        if commit in selected_commits:
+            yield tag, commit
 
 def _get_release_commits(repo, release_line):
+    """Get the commits for a release line."""
     match = re.match(RELEASE_LINE_PATTERN, release_line)
 
     if not match:
         raise ValueError(f"Invalid release line specification: {release_line}")
 
     rev = match.groupdict()["rev"] or repo.get_current_branch()
+    
+    if match.groupdict()["rev"] == "_":
+        # loop over all remote branches and yield the tags for the commits
+        
+        tags = set()
+        for branch in repo.refs.as_dict(b"refs/remotes/origin/").keys():
+            rev = branch.decode()                
+            rev_selection = match.groupdict()["rev_selection"]
 
-    if match.groupdict()["rev_selection"] is None:
+            for tag, commit in _get_tags(repo, rev, rev_selection):
+                if tag not in tags:
+                    tags.add(tag)
+                    yield tag, commit
+                
+        return
+
+    elif match.groupdict()["rev_selection"] is None:
         # No rev_selection means to select this and only this specific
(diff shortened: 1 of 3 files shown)


unkcpz:
> Wouldn't a start (*) be better?

Sure, it is better I think, the standard wildcard.

aiidalab/registry/releases.py
```diff
@@ -30,15 +30,62 @@ def _split_release_line(url):
     return url, None
 
 
+def _get_tags(repo, rev, rev_selection):
```
danielhollas, line 33: This is a non-trivial function, could you please add a docstring? Especially explaining the input parameters and the output. Adding types would help understanding as well.

aiidalab/registry/releases.py
```diff
@@ -30,15 +30,62 @@ def _split_release_line(url):
...
+    selected_commits = repo.rev_list(f"{start or ref}..{stop or ref}")
+
+    for tag in repo.get_merged_tags(rev):
+        commit = repo.get_commit_for_tag(tag)
+        if commit in selected_commits:
+            yield tag, commit
+
+
 def _get_release_commits(repo, release_line):
+    """Get the commits for a release line."""
```
danielhollas, line 65: It would be nice to explain what is the "release_line", unless this is already documented somewhere up the stack.
Also please add types.

aiidalab/registry/releases.py
```diff
@@ -30,15 +30,62 @@ def _split_release_line(url):
...
+    if match.groupdict()["rev"] == "*":
+        # loop over all remote branches and yield the tags for the commits
+
+        tags = set()
+        for branch in repo.refs.as_dict(b"refs/remotes/origin/").keys():
+            rev = branch.decode()
+            rev_selection = match.groupdict()["rev_selection"]
+
+            for tag, commit in _get_tags(repo, rev, rev_selection):
+                if tag not in tags:
```
danielhollas, line 82: Just so understand, this means that if the developer borks up and has the same tags in different branches, this will pick just one of them, essentially at random? I am wondering if we should rather fail in that case? (it would probably be quite disruptive, but just ignoring this also seems potentially problematic).

aiidalab/registry/releases.py
```diff
@@ -30,15 +30,62 @@ def _split_release_line(url):
...
     match = re.match(RELEASE_LINE_PATTERN, release_line)
 
     if not match:
         raise ValueError(f"Invalid release line specification: {release_line}")
 
     rev = match.groupdict()["rev"] or repo.get_current_branch()
 
-    if match.groupdict()["rev_selection"] is None:
+    if match.groupdict()["rev"] == "*":
+        # loop over all remote branches and yield the tags for the commits
```
danielhollas, line 74: Just wondering, is this potentially an expensive operation to iterate over all branches? Hopefully not, but it would be good to somehow test / verify.

tests/test_registry.py
```diff
@@ -0,0 +1,34 @@
...
+
+
+def test_get_all_tagged_releases():
+    """Test that all tagged releases are returned."""
+    url = "git+https://github.com/aiidalab/aiidalab-qe.git@*:v23.04.0^.."
+    base_url, release_line = _split_release_line(url)
+    parsed_url = urlsplit(base_url)
+
+    with fetch_from_url(base_url) as repo_path:
+        if parsed_url.scheme.startswith("git+"):
```
danielhollas, line 15: Surely this if is not needed when we have the URL defined above?

tests/test_registry.py
```diff
@@ -0,0 +1,34 @@
+import os
+from urllib.parse import urlsplit
+
+from aiidalab.fetch import GitRepo, fetch_from_url
+from aiidalab.registry.releases import _get_release_commits, _split_release_line
+
+
+def test_get_all_tagged_releases():
+    """Test that all tagged releases are returned."""
+    url = "git+https://github.com/aiidalab/aiidalab-qe.git@*:v23.04.0^.."
```
danielhollas, line 10: Thanks for adding the test! :clap:  
It's not great that we rely here on an external URL, but at least we'll find out if we ever deleted that branch by accident. :D

danielhollas (changes requested):
Thanks @unkcpz, mostly nitpicks and docstrings/types.

Also it would be great to have @yakutovicha's blessing that he's happy with this approach.

aiidalab/registry/releases.py
unkcpz: It is not, it runs pretty fast. I run `build.sh` in aiidalab-regstry locally and I cannot feel any delay compared to only fetching on one branch. In QeApp, we have quite a lot branches (this is not good 🥲 ) but it just quickly finished without I noticing it.

aiidalab/registry/releases.py
unkcpz: I don't think it is possible to have two commits with the same tag, right? 
This code is quite redundant, in principle the tags should get like:

```python
repo = Repo(repo_path)
tags = repo.refs.as_dict("refs/tags")
```

But this works not well with the rev_list we already introduced and supported. Thus I decided to loop over all branches with rev_list supported.

tests/test_registry.py
unkcpz: Yep! Removed.

aiidalab/registry/releases.py
unkcpz: I change `rev` to `branch` as the check happened in the function.

aiidalab/registry/releases.py
danielhollas: You are corrent, I just tested that git does not allow you to create tags with the same name for different commits. Thanks!

danielhollas (approved):
Cool, let's try it out!

PR #387, opened 2023-06-27
Update manifest

danielhollas:
Just removing non-existent files from the manifest


unkcpz (approved):
Looks all good to me! Thanks.

PR #388, opened 2023-06-27
Unpin MarkupSafe package

danielhollas:
The pin was a workaround around an issue with an old Jinja 2.x version, but we have since then updated Jinja so it should not be neccessary anymore.

See https://github.com/aiidalab/aiidalab/pull/264 for details.

NOTE: This pin also prevents potential upgrade to Python 3.11

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -21,8 +21,7 @@ classifiers =
 packages = find:
 install_requires =
     CacheControl~=0.12
-    Jinja2>=2.11.3,<4
-    MarkupSafe==2.0.1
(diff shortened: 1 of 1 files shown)


danielhollas:
I've tested that the latest MarkupSafe works with `jinja2==3.1.2` that is installed in the current Docker image.

unkcpz (commented):
Is `jinja2` version need to update as well?  If the `Jinja2==2.11.3` is installed, the problem is still there?

unkcpz (approved):
Great, thanks!

PR #390, opened 2023-07-10
DOC: aiidalab-launch CLI using profile to align with verdi command

unkcpz:


yakutovicha (approved):
I think it is quite safe to go with this change, please go ahead @unkcpz 👍

Issue #363, opened 2023-03-03
Complete the mypy check for whole package


unkcpz:
The typing hint check only happened on the functions we encountered in https://github.com/aiidalab/aiidalab/pull/347
There still files exclude from mypy check and using `#mype: ignore` to suppress the warning. This issue is open until all mypy check is passed.

ayushjariyal:
Hey @unkcpz , I want to try this. Can you please show me the reference for the files that have not been typed yet?

danielhollas:
Hi @ayushjariyal,

most of the typing have been done here, but if you want you can still search for/grep for `type: ignore` statements or `Any` types throughout the repo.

You can also review #454, which is a last PR I did related to typing, after which I'll close this issue.

PR #391, opened 2023-08-07
Add typing to git_util.py,  metadata.py and config.py

danielhollas:
Step towards #363. I am definitely not a typing expert, taking this work as an opportunity to learn. Would appreciate help to get rid of the `type: ignore`s (although some of them seems to be caused by `dulwich` package not providing types).

### DIFF
--- a/aiidalab/config.py
+++ b/aiidalab/config.py
@@ -1,6 +1,7 @@
 """Module to manage AiiDAlab configuration."""
 from os import getenv
 from pathlib import Path
+from typing import Any
 
 import click
 import toml
@@ -18,11 +19,11 @@ if _DEVELOP_MODE:  # Warn developer that the mode is enabled.
     click.secho("\n".join([f"\U0001F6A7  {line}" for line in lines]), fg="yellow")
 
 
-def _as_env_var_name(key):
+def _as_env_var_name(key: str) -> str:
     return "AIIDALAB_" + key.upper()
 
 
-def _get_config_value(key, default=None):
+def _get_config_value(key: str, default: str) -> Any:
     """Return config value from configuration source.
 
     The standard configuration source order is:
--- a/aiidalab/config.py
+++ b/aiidalab/config.py
@@ -1,7 +1,7 @@
 """Module to manage AiiDAlab configuration."""
 from os import getenv
 from pathlib import Path
-from typing import Any
+from typing import Any, Optional
 
 import click
 import toml
@@ -23,7 +23,7 @@ def _as_env_var_name(key: str) -> str:
     return "AIIDALAB_" + key.upper()
 
 
-def _get_config_value(key: str, default: str) -> Any:
+def _get_config_value(key: str, default: Optional[str] = None) -> Any:
     """Return config value from configuration source.
 
     The standard configuration source order is:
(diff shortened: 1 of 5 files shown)


aiidalab/config.py
```diff
@@ -18,11 +19,11 @@
     click.secho("\n".join([f"\U0001F6A7  {line}" for line in lines]), fg="yellow")
 
 
-def _as_env_var_name(key):
+def _as_env_var_name(key: str) -> str:
     return "AIIDALAB_" + key.upper()
 
 
-def _get_config_value(key, default=None):
+def _get_config_value(key: str, default: str) -> Any:
```
unkcpz, line 26:
```suggestion
def _get_config_value(key: str, default: str | None) -> Any:
```

`default` is passed to `os.getenv` and accept to be `None`.

aiidalab/git_util.py
```diff
@@ -223,10 +237,10 @@ def get_current_branch(self):
             )
         return branch.strip()
 
-    def get_commit_for_tag(self, tag):
+    def get_commit_for_tag(self, tag: str):  # type: ignore
```
unkcpz, line 240: Not sure why here the type ignore required, this function return value, possible to add the type for `self.get_peeled`?

aiidalab/git_util.py
```diff
@@ -120,22 +134,22 @@ def git_clone(url, commit, path):
 
 
 @dataclass
-class GitPath(os.PathLike):
+class GitPath(os.PathLike):  # type: ignore
```
unkcpz, line 137: We may want to avoid to use ignore for subclassing by setting https://mypy.readthedocs.io/en/stable/command_line.html#cmdoption-mypy-disallow-subclassing-any

aiidalab/git_util.py
```diff
@@ -21,14 +35,14 @@ class BranchTrackingStatus(Enum):
     DIVERGED = 2
 
 
-class GitManagedAppRepo(Repo):
+class GitManagedAppRepo(Repo):  # type: ignore
     """Utility class to simplify management of git-based apps."""
 
-    def list_branches(self):
+    def list_branches(self):  # type: ignore
```
unkcpz, line 41:
```suggestion
    def list_branches(self) -> Any: 
```
Not sure this is good, we know it has a return value, so maybe just return Any?

unkcpz (changes requested):
@danielhollas thanks! Just minor requests. I don't think there is too many `#type: ignore`, as you said they are mostly from other libraries.

aiidalab/git_util.py
danielhollas: I think `get_peeled` is inherited from `dulwich`. I've added `Any` as return type. Looks like `decode` function returns Any, so not sure we can do better

aiidalab/metadata.py
```diff
@@ -1,15 +1,17 @@
 """App metadata specification"""
 from __future__ import annotations
 
-from configparser import ConfigParser
+from configparser import ConfigParser, SectionProxy
```
unkcpz, line 4:
```suggestion
from typing import TYPE_CHECKING

if TYPE_CHECKING:
    from configparser import ConfigParser, SectionProxy
    from pathlib import Path
```

A minor nitpick, you are the expert on module loading performance, will this improve the loading efficiency?

unkcpz (approved):

aiidalab/metadata.py
danielhollas: Not really, both of these are stdlib modules, which are generally very fast to import. More importantly, `pathlib` is getting imported all over the place anyway, and you would anyway need to import `ConfigParser` here.

PR #392, opened 2023-08-08
Add typing to fetch.py, more types in app.py

danielhollas:
Besides `fetch.py` I started working on `app.py` but quickly realized it will be more work than I had time for now so I only did the most obvious stuff for now. This allowed me to delete some skipped errors in setup.cfg.

### DIFF
--- a/aiidalab/app.py
+++ b/aiidalab/app.py
@@ -682,7 +682,7 @@ class AiidaLabAppWatch:
         if self._monitor_thread is not None:
             self._monitor_thread_stop.set()
 
-    def is_alive(self) -> Any:
+    def is_alive(self) -> bool | None:
         """Return True if this watch is still alive."""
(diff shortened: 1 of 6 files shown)


aiidalab/app.py
```diff
@@ -667,16 +677,16 @@ def check_path_exists_changed():
...
         self._started = True
 
-    def stop(self):
+    def stop(self) -> None:
         """Stop watching the app repository for file system events."""
         if self._monitor_thread is not None:
             self._monitor_thread_stop.set()
 
-    def is_alive(self):
+    def is_alive(self) -> Any:
```
danielhollas, line 685: I tried using `bool` here but mypy complained. I don't understand why, I would think that the `and` operator always returns bool?

danielhollas: Ooh, interesting. `bool | None` seems to work in this case.

PR #395, opened 2023-09-11
Move developer guide to Wiki

unkcpz:
As we discussed, we want to leave the docs of aiidalab/aiidalab for the users and move the documentation of development guides for developers to GitHub repo wiki (https://github.com/aiidalab/aiidalab/wiki).


danielhollas:
@unkcpz can you please enable the readthedocs build on PRs on this repo? (same way as is done in AWB repo)

unkcpz:
In is enabled https://aiidalab--395.org.readthedocs.build/en/395/

danielhollas (approved):
Thanks! LGTM except for one minor correction.

I would also personally remove the 'API documentation` link in the header, since again that is not something that is useful to the user and is potentially confusing. But that can be done in a separate PR if you agree.

unkcpz:
> I would also personally remove the 'API documentation` link in the header, since again that is not something that is useful to the user and is potentially confusing.

I agree, I simply remove it. For the moment, I didn't find any case that is useful.

yakutovicha:
> > I would also personally remove the 'API documentation` link in the header, since again that is not something that is useful to the user and is potentially confusing.
> 
> I agree, I simply remove it in [3559e73](https://github.com/aiidalab/aiidalab/commit/3559e73a13188d2e0351686509958825dd4cc910). For the moment, I didn't find any case that is useful. Let me know if you agree with it.

I would strongly advocate for retaining the API link. This is essential for the development of the tools that rely on aiidalab. I recall that we reached a consensus to transition only the internal development-related content to the wiki.

unkcpz:
> I would strongly advocate for retaining the API link.

That's true.. It is just a bit confusing the hyperlink is on the header. Maybe I can keep it but remove it from the header and keep it in the outline. So the developers still can find it on the doc page. I'll make this change and we see if we can agree on a format.

danielhollas:
> That's true.. It is just a bit confusing the hyperlink is on the header. Maybe I can keep it but remove it from the header and keep it in the outline. So the developers still can find it on the doc page. I'll make this change and we see if we can agree on a format.

That seems like a good compromise. My concern was indeed that we're mixing user docs with develop docs. And note that these docs are about AiiDAlab in general, not really about the `aiidalab` package itself which is why it is confusing. Let's keep the link somewhere (could also be in the README repo and on the Wiki) but not in the header. WDYT @yakutovicha ?

yakutovicha:
> > That's true.. It is just a bit confusing the hyperlink is on the header. Maybe I can keep it but remove it from the header and keep it in the outline. So the developers still can find it on the doc page. I'll make this change and we see if we can agree on a format.
> 
> That seems like a good compromise. My concern was indeed that we're mixing user docs with develop docs. And note that these docs are about AiiDAlab in general, not really about the `aiidalab` package itself which is why it is confusing. Let's keep the link somewhere (could also be in the README repo and on the Wiki) but not in the header. WDYT @yakutovicha ?

In my opinion, we should be guided by the practicality.

To allow quick referencing to the code we should include the API in the documentation.

Combining user and developer documentation isn't an issue as long as they remain categorized; it only becomes problematic if they become intertwined.

It also makes sense to keep AiiDAlab docs with the package docs since they're closely related, which makes documentation management easier and reduces the need for lots of cross-referencing between separate sources.

The internal development things are better to be in the wiki because they have a very low entrance barrier and allow for quick communication/editing.

yakutovicha:
@unkcpz here is for the doc build fix: https://github.com/aiidalab/aiidalab-widgets-base/pull/373/commits/523e20775d114727bc6856acc4953a6605b56b60

unkcpz:
Thanks @yakutovicha. the doc now build success but I don't know how to remove the API doc from the header. Almost give up...

unkcpz:
Also add discourse link and update the `pydata-theme` version

docs/source/conf.py
```diff
@@ -40,7 +39,7 @@
 
 # General information about the project.
 project = "aiidalab"
-copyright_first_year = "2020"
```
yakutovicha, line 43: I think this should remain `2020`. It is the "first year" not the current year.

yakutovicha (changes requested):
one tiny request, otherwise good to go 👍

docs/source/conf.py
```diff
@@ -40,7 +39,7 @@
...
+copyright_first_year = "2023"
```
unkcpz, line 42:
```suggestion
copyright_first_year = "2020"
```

docs/source/conf.py
unkcpz: Ah, yes. I changed it back.

yakutovicha (approved):
LGTM!

PR #402, opened 2024-01-31
Pre-commit update + fix docs build

danielhollas:
Update of mypy now detects some new issues. Not sure if these are real or a mypy bug, need to investigate more.

EDIT: Looks like mypy was right and we were not ensuring that we return a proper types in one of the functions. The refactor I made is not very pretty but does the trick. Further cleanup would be more work I think.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -4,7 +4,7 @@ ci:
 
 repos:
     - repo: https://github.com/pre-commit/pre-commit-hooks
-      rev: v4.4.0
+      rev: v4.5.0
       hooks:
           - id: check-json
           - id: check-yaml
@@ -23,7 +23,7 @@ repos:
           - id: yamlfmt
 
     - repo: https://github.com/psf/black
-      rev: 23.9.1
+      rev: 23.12.1
       hooks:
           - id: black
             language_version: python3 # Should be a command that runs python3.6+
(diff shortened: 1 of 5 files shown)


unkcpz:
For the doc build failure, I guess https://github.com/aiidateam/aiida-core/pull/6253 this will help? Or maybe just simply update the sphinx version to the latest.

danielhollas:
I've managed to fix the docs build in the end (in an ugly way). LMK if you want me to split it to a separate PR.

yakutovicha:
> I've managed to fix the docs build in the end (in an ugly way). LMK if you want me to split it to a separate PR.


Since fixing Sphinx problems requires adding many dependencies, why don't we switch to a later version of Sphinx itself? Are there any issues with such an update??

yakutovicha:
Also, the latest CI test failed because we are still on Python 3.8. Time to go on with Python>3.9?

unkcpz:
> Time to go on with Python>3.9?

I think so, we drop it in AWB https://github.com/aiidalab/aiidalab-widgets-base/pull/557

danielhollas:
Sorry for not being clear, I explain the reason why we can't simply update in a note in `setup.cfg`
https://github.com/aiidalab/aiidalab/pull/402/files#diff-fa602a8a75dc9dcc92261bac5f533c2a85e34fcceaff63b3a3a81d9acde2fc52R64

IMO the workaround is okay for now, but of course we need to fix it eventually.

> Also, the latest CI test failed because we are still supporting Python 3.8. Time to go on with Python>3.9?

This PR has already more stuff than I planned so I'll leave that to the next PR.
For now I simply put pre-commit v3.5 which is still compatible with Python 3.8.

yakutovicha:
> > Since fixing Sphinx problems requires adding many dependencies, why don't we switch to a later version of Sphinx itself? Are there any issues with such an update??
> 
> Sorry for not being clear, I explain the reason why we can't simply update in a note in `setup.cfg` https://github.com/aiidalab/aiidalab/pull/402/files#diff-fa602a8a75dc9dcc92261bac5f533c2a85e34fcceaff63b3a3a81d9acde2fc52R64
> 
> Note that we're not adding new dependencies, we're pinning transitive dependencies that were already there. IMO the workaround is okay for now, but of course we need to fix it eventually.
> 
> > Also, the latest CI test failed because we are still supporting Python 3.8. Time to go on with Python>3.9?
> 
> This PR has already more stuff than I planned so I'll leave that to the next PR. For now I simply put pre-commit v3.5 which is still compatible with Python 3.8.

Thanks for explaining @danielhollas, makes sense to me.

PR #414, opened 2024-02-23
Group Dependabot PRs

danielhollas:
Let's try config so that dependabot does not open separate PR for every individual GH action.


unkcpz (approved):
Yep, let's try. As I know, there is no way to trigger it manually. But also not to much to test, it is pretty standard thing I assume. I use it in https://github.com/unkcpz/aiida-core-i18n

PR #419, opened 2024-03-12
Use uv to install dependencies in CI

danielhollas:
[uv](https://github.com/astral-sh/uv/issues/2155l) is the new cool kid in the town of Python packaging, from the creators of ruff. It currently serves as a much faster drop-in replacement for pip. 

For the `aiidalab` package, this drops installation time from ~15s to ~2s, which nearly halves the workflow time since the tests we have here are pretty quick.

Couple other small changes (feel free to push back)
 - Remove pre-commit section from CI since we use the pre-commit.ci anyway. 
 - Don't run tests on both push and pull_request event, this leads to duplication for pull requests coming from origin.
 - ~~remove bumpver and pre-commit from direct dev dependencies, as proposed [on Discourse](https://aiida.discourse.group/t/remove-pre-commit-and-bumpver-from-dev-dependencies/325)~~
 - get rid of incidental pyyaml dependency, see comment below

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -7,6 +7,9 @@ on:
             - main
     pull_request:
 
+env:
+    FORCE_COLOR: 1
+
 jobs:
 
     test-package:
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -29,7 +29,7 @@ jobs:
                   python-version: ${{ matrix.python-version }}
 
             - name: Install uv
-              run: curl -LsSf https://astral.sh/uv/install.sh | sh
+              run: curl --proto '=https' --tlsv1.2 -LsSf https://github.com/astral-sh/uv/releases/download/0.1.21/uv-installer.sh | sh
 
             - name: Install package with uv
               run: uv pip install --system -e .[dev]
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -29,7 +29,7 @@ jobs:
                   python-version: ${{ matrix.python-version }}
 
             - name: Install uv
-              run: curl --proto '=https' --tlsv1.2 -LsSf https://github.com/astral-sh/uv/releases/download/0.1.21/uv-installer.sh | sh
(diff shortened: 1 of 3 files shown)


setup.cfg
```diff
@@ -51,8 +51,6 @@ console_scripts =
 
 [options.extras_require]
 dev =
-    bumpver==2021.1114
```
danielhollas, line 54: Unrelated change, I asked about this on Discourse.

tests/conftest.py
```diff
@@ -1,7 +1,7 @@
 from pathlib import Path
 
 import pytest
-import yaml
+from ruamel.yaml import YAML
```
danielhollas, line 4: We've been depending on `pyyaml` for tests by accident, and did not notice since it got installed together with pre-commit. Let's use ruamel since that's what is used in the aiidalab package.

.github/workflows/ci.yml
```diff
@@ -36,10 +28,11 @@ jobs:
               with:
                   python-version: ${{ matrix.python-version }}
 
-            - name: Install package
-              run: |
-                  pip install -e .[dev]
-                  pip freeze
+            - name: Install uv
+              run: curl --proto '=https' --tlsv1.2 -LsSf https://github.com/astral-sh/uv/releases/download/0.1.21/uv-installer.sh | sh
```
danielhollas, line 32:
```suggestion
              run: curl --proto '=https' --tlsv1.2 -LsSf https://github.com/astral-sh/uv/releases/download/0.1.31/uv-installer.sh | sh
```

danielhollas (commented):
@superstar54 could you take a look? This might be beneficial for QeApp as well.

setup.cfg
unkcpz: Hi @danielhollas, you add it back in the new commit correct?

danielhollas: I did, since we didn't yet resolve this conversation on Discourse. (I still think it would be beneficial to remove, I'll reply on Discourse when I have time)

unkcpz: Another option maybe don't pin down the version then if it already installed it will use it, otherwise install the latest version.

danielhollas: I've unpinned them for now.

danielhollas:
Going to merge this, we've got good experience with uv in AWB.

PR #422, opened 2024-04-03
Switch to ruff linter and formatter

danielhollas:
I've taken ruff config from AWB (with minor modifications as needed).

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -5,7 +5,7 @@ updates:
     - package-ecosystem: github-actions
       directory: /
       schedule:
-          interval: daily
+          interval: monthly
       groups:
           gha-dependencies:
               patterns:
--- a/aiidalab/app.py
+++ b/aiidalab/app.py
@@ -475,7 +475,7 @@ class _AiidaLabApp:
                         for member in tar.getmembers():
                             member_path = os.path.join(path, member.name)
                             if not is_within_directory(path, member_path):
-                                raise Exception("Attempted Path Traversal in Tar File")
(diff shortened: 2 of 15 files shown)


.github/dependabot.yml
```diff
@@ -5,7 +5,7 @@ updates:
     - package-ecosystem: github-actions
       directory: /
       schedule:
-          interval: daily
+          interval: monthly
```
danielhollas, line 8: This seems a bit saner and matches what we have in AWB.

aiidalab/app.py
```diff
@@ -569,15 +569,15 @@ def install(
                         "Performing rollback to previously installed version."
                     )
                     self._restore_from(trash_path)
-            except RuntimeError as inner_error:
-                logger.error(f"Rollback failed due to error: {inner_error}!")
+            except RuntimeError:
+                logger.exception("Rollback failed! Consider re-installing the app.")
```
danielhollas, line 573: Apparently, the exception is always automatically included when using `logger.exception` inside except statement.

aiidalab/registry/web.py
```diff
@@ -121,7 +122,6 @@ def build(
                 api_path=base_path / api_path,
                 apps_index=apps_index,
                 apps_data=apps_data,
-                scan_app_repository=parse_app_repo,
```
danielhollas, line 124: This argument was completely unused in `build_api_v1`, perhaps copy-paste error or some historical remnant?

yakutovicha (approved):
LGTM!

PR #428, opened 2024-06-21
Remove ipython as a direct dependency

danielhollas:
The `aiidalab` package depends transitively on ipython via `traitlets`, but we do not use it directly in the package so we shouldn't pin its version in the package requirements.

This is currently a blocker for updating the base jupyter image to python 3.10/3.11 in https://github.com/aiidalab/aiidalab-docker-stack/pull/455


unkcpz (approved):
Make sense to delete it, thanks! @danielhollas

PR #431, opened 2024-06-22
Add basic CLI tests

danielhollas:
Add basic smoke tests for aiidalab CLI, such as running `aiidalab info` etc. 

~~The overall coverage in Codecov is down overall, because we were not tracking coverage of `aiidalab/__main__.py`.~~ This PR bumps code coverage by 7%.

Split from #429


unkcpz:
Shouldn't we use `CliRunner` instead?

danielhollas:
> Shouldn't we use CliRunner instead?

@unkcpz indeed, I was a bit lazy, thanks for the reminder! I've converted the new tests to use it.

danielhollas:
I've gone ahead and merged this since these were mostly test changes. Happy to tweak this in a follow-up PR.

PR #438, opened 2024-07-13
Support python 3.12

danielhollas:
Bumped watchdog dependency to support python 3.12
According to changelog there shouldn't be any breaking changes. 🤞


unkcpz (approved):
Thanks!

PR #439, opened 2024-07-13
Drop support for old docker stack and Python-3.8

danielhollas:
Drop support for the old docker stack, which is no longer maintained (and hopefully not used anywhere!). We can then drop support for Python 3.8 since the new stack started with 3.9

This reverts relevant parts of #355.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -19,7 +19,7 @@ jobs:
         strategy:
             fail-fast: false
             matrix:
-                python-version: ['3.8', '3.9', '3.10', '3.11']
+                python-version: ['3.9', '3.11', '3.12']
 
         steps:
 
--- a/aiidalab/app.py
+++ b/aiidalab/app.py
@@ -42,7 +42,6 @@ from aiidalab.utils import (
     load_app_registry_index,
     run_pip_install,
     run_post_install_script,
-    run_reentry_scan,
     run_verdi_daemon_restart,
     split_git_url,
     this_or_only_subdir,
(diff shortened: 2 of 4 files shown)


.github/workflows/ci.yml
```diff
@@ -19,7 +19,7 @@ jobs:
         strategy:
             fail-fast: false
             matrix:
-                python-version: ['3.8', '3.9', '3.10', '3.11']
+                python-version: ['3.9', '3.11', '3.12']
```
danielhollas, line 22: We'll skip Python 3.10 in the docker stack so it is kind of pointless to test with it.

aiidalab/app.py
```diff
@@ -393,23 +378,14 @@ def _pip_install(*args, stdout):
...
-                    logger.warning("'reentry scan' failed")
-
             # Restarting the AiiDA daemon to import newly installed plugins.
+            # TODO: Skip this if verdi is not available (useful for testing).
             process = run_verdi_daemon_restart()
             for line in io.TextIOWrapper(process.stdout, encoding="utf-8"):
                 stdout.write(line)
             process.wait()
             if process.returncode != 0:
+                # TODO: I don't think we should fail the installation if this step fails.
```
unkcpz, line 388: It is a bit safe to fail or at least a big warning alert. Because I think restart the daemon here is mainly to make plugins available by daemon. 
It will be a bit confuse for user if they install app but not able to restart daemon, and then they run calculation and get error like "module xxx is not found".

danielhollas: Yeah, I am a bit torn on this. The disadvantage here is that one cannot install apps when the daemon is not running, which complicates testing, and also does not make sense conceptually.

In any case this is just a comment, I do not plan to change anything in this PR.

unkcpz: > The disadvantage here is that one cannot install apps when the daemon is not running

Maybe just check if it is not running, using `start`? It calls `run_verdi_daemon_restart`, so I guess it has enough flexibility to put a bit logic there?

danielhollas: I created #441 to follow up on this.

PR #440, opened 2024-07-13
Add `aiidalab install` test

danielhollas:
Added a couple of simple CLI tests for the `aiidalab install` command + some supporting infrastructure and small fixes.

More installation tests are coming in a separate PR to keep the changes small.

### DIFF
--- a/aiidalab/__main__.py
+++ b/aiidalab/__main__.py
@@ -470,7 +470,12 @@ def install(
     is_flag=True,
     help="Ignore all warnings and perform potentially dangerous operations anyways.",
 )
-def uninstall(app_name, yes, dry_run, force):
+@click.option(
+    "--fully-remove",
+    is_flag=True,
+    help="Do not move application directory to ~/.trash.",
+)
+def uninstall(app_name, yes, dry_run, force, fully_remove):
     """Uninstall apps."""
     from .config import AIIDALAB_APPS
 
@@ -545,7 +550,7 @@ def uninstall(app_name, yes, dry_run, force):
                         f"Would uninstall '{app.name}' ('{app.path!s}').", err=True
                     )
                 else:
-                    app.uninstall()
+                    app.uninstall(move_to_trash=not fully_remove)
                     click.echo(
                         f"Uninstalled '{app.name}' ('{app.path!s}').",
                         err=True,
--- a/aiidalab/__main__.py
+++ b/aiidalab/__main__.py
@@ -433,28 +433,26 @@ def install(
         ],
         headers=["App", "Version", "Path"],
     )
-    click.echo(f"Would install:\n\n{indent(apps_table, '  ')}\n")
+    click.echo(f"Installation plan:\n\n{indent(apps_table, '  ')}\n")
     if yes or click.confirm("Proceed?", default=True):
         for app, version in install_candidates.values():
-            if version is not None:
-                if dry_run:
-                    click.secho(
-                        f"Would install '{app.name}' version '{version}'.", fg="green"
-                    )
-                else:
-                    try:
-                        app.install(
-                            version=version,
-                            install_dependencies=dependencies == "install",
-                            python_bin=python_bin,
-                        )
-                    except RuntimeError as error:
-                        click.secho(f"Error: {error}", fg="red")
-                        break
-                    else:
-                        click.secho(
-                            f"Installed '{app.name}' version '{version}'.", fg="green"
-                        )
+            if version is None:
+                continue
+            if dry_run:
+                click.secho(
+                    f"Would install '{app.name}' version '{version}'.", fg="green"
+                )
+                continue
+            try:
+                app.install(
+                    version=version,
+                    install_dependencies=dependencies == "install",
+                    python_bin=python_bin,
+                )
+            except RuntimeError as error:
+                raise click.ClickException(error)
+            else:
+                click.secho(f"Installed '{app.name}' version '{version}'.", fg="green")
 
 
 @cli.command()
--- a/aiidalab/__main__.py
+++ b/aiidalab/__main__.py
@@ -473,6 +473,7 @@ def install(
 @click.option(
     "--fully-remove",
     is_flag=True,
+    hidden=True,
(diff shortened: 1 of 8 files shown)


aiidalab/__main__.py
```diff
@@ -433,28 +433,26 @@ def install(
...
-                            python_bin=python_bin,
-                        )
-                    except RuntimeError as error:
-                        click.secho(f"Error: {error}", fg="red")
-                        break
-                    else:
-                        click.secho(
-                            f"Installed '{app.name}' version '{version}'.", fg="green"
-                        )
+            if version is None:
```
danielhollas, line 439: Using this 'early-continue' pattern to reduce indentation.

aiidalab/__main__.py
```diff
@@ -433,28 +433,26 @@ def install(
...
+                )
+                continue
+            try:
+                app.install(
+                    version=version,
+                    install_dependencies=dependencies == "install",
+                    python_bin=python_bin,
+                )
+            except RuntimeError as error:
+                raise click.ClickException(error)
```
danielhollas, line 453: This is a small bugfix so that we return non-zero exit code when there's error.

aiidalab/app.py
```diff
@@ -49,7 +47,6 @@
     throttled,
 )
 
-from .config import AIIDALAB_APPS
```
danielhollas, line 52: Need to move this import otherwise AIIDALAB_APPS won't be patched in tests.

aiidalab/app.py
```diff
@@ -248,7 +250,7 @@ def _move_to_trash(self) -> Path | None:
         trash_path = Path.home().joinpath(".trash", f"{self.name}-{uuid4()!s}")
         if self.path.exists():
             trash_path.parent.mkdir(parents=True, exist_ok=True)
-            self.path.rename(trash_path)
```
danielhollas, line 251: This would fail with `OSError`  if the two directories were on different filesystems. (such as `/home/` and `/tmp`). `shutil.move` handles this case automatically.

aiidalab/app.py
```diff
@@ -572,9 +578,9 @@ def install(
             except RuntimeError:
                 logger.exception("Rollback failed! Consider re-installing the app.")
             finally:
-                raise RuntimeError(
-                    f"Failed to install '{self.name}' (version={version}) at '{self.path}'"
-                ) from error
+                logger.info("Full stack trace", exc_info=True)
```
danielhollas, line 581: This displays the full stack trace when a larger verbosity is used (`aiidalab -v install`)

aiidalab/environment.py
```diff
@@ -46,8 +46,6 @@ class Environment:
 
     python_requirements: list[str] = field(default_factory=list)
 
-    _FILES = ("requirements.txt",)
```
danielhollas, line 49: Doesn't seem to be used anywhere.

tests/conftest.py
```diff
@@ -62,6 +62,43 @@ def _generate_app(
 ]
 
 
+@pytest.fixture(autouse=True, scope="function")
+def forbid_external_commands(monkeypatch):
```
danielhollas, line 66: NOTE: these fixtures actually don't work for the CLI tests in this PR since they are run via Runner which launches `aiidalab` as external process. But they will be useful later for unit-style tests.

unkcpz:
This sounds cool. Is this mainly target for developer or it targets to integrate to GUI?
I remember we had a discussion to make install two steps, first install to some sandbox and then copy to home? 

I didn't take close look, but how you manage the dependencies? Is the dependencies also cleaned up if dry-run failed?

danielhollas:
> This sounds cool. Is this mainly target for developer or it targets to integrate to GUI?

I am not sure I understand the question, this PR is about testing, it doesn't add any new features. Sorry if the PR description was not clear. The `--dry-run` option already existed, it doesn't actually install anything, it mostly just validates that the package exists I think.

unkcpz:
Ah, I see. This PR mainly bring tests to command.

danielhollas:
@unkcpz will you have time to review this sometimes this week? I'll need to these tests for the other PRs. It's okay if not, I'll merge this and am happy to take feedback after when you have time. Thanks!

aiidalab/__main__.py
```diff
@@ -433,28 +433,27 @@ def install(
...
-                else:
-                    try:
-                        app.install(
-                            version=version,
-                            install_dependencies=dependencies == "install",
-                            python_bin=python_bin,
-                        )
-                    except RuntimeError as error:
-                        click.secho(f"Error: {error}", fg="red")
-                        break
```
unkcpz, line 453: Well, not exactly the same by changing to raise exception if there are more things going after the for loop ;)

aiidalab/__main__.py
```diff
@@ -472,7 +471,13 @@ def install(
     is_flag=True,
     help="Ignore all warnings and perform potentially dangerous operations anyways.",
 )
-def uninstall(app_name, yes, dry_run, force):
+@click.option(
+    "--fully-remove",
```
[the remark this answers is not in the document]
unkcpz: For a good naming, maybe use `--move-to-trash`, revert the bool and set default to `true`?

tests/conftest.py
unkcpz: I am not quite understand how this would work. If the unit test is running with this fixture loaded, will it monitor the system call of `pip install` etc?

unkcpz (approved):
Thanks @danielhollas, tests are neat and fancy!

tests/conftest.py
danielhollas: In this fixture I am using `monkeypatch` to replace the python functions in `aiidalab.utils` module. If the unit tests ends up calling them, this fixture will raise an exception. 

I'll see if this makes sense once I start writing the tests, I will change it if not.

aiidalab/__main__.py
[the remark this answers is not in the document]
danielhollas: I agree, I'll change it in a future PR.

PR #443, opened 2024-07-23
Make auto-generated release notes nicer

danielhollas:
Adds a configuration for auto-generated release notes to make them a bit nicer:
- Exclude PRs from dependabot and pre-commit-ci bot
- Autogenerate sections based on PR labels.

I've already merged tested this in AWB https://github.com/aiidalab/aiidalab-widgets-base/pull/618 and it seems to work nicely, let's try it here as well.

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -94,7 +94,7 @@ jobs:
                   name: release
                   path: dist/
 
-            - uses: softprops/action-gh-release@v2.0.6
+            - uses: softprops/action-gh-release@v2
               name: Create release
               if: startsWith(github.ref, 'refs/tags/v')
               with:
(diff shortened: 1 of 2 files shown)


.github/workflows/publish.yml
```diff
@@ -94,7 +94,7 @@ jobs:
                   name: release
                   path: dist/
 
-            - uses: softprops/action-gh-release@v2.0.6
+            - uses: softprops/action-gh-release@v2
```
danielhollas, line 97: I've unpinned the softprops/actions-gh-release action.
Pinning this action to specific patch version just creates noise from dependabot. Given that we currently don't pin any other actions to the exact versions, let's unpin this one as well.

danielhollas: Same as in https://github.com/aiidalab/aiidalab-docker-stack/pull/489

unkcpz (approved):
Thanks! @danielhollas

PR #447, opened 2024-08-09
Fix segfault from requests_cache

danielhollas:
As #196 explains, we now use a global cache for *all* `requests` calls! That's quite scary since it influences apps that import `aiidalab` package, such as AWB or home app! 

Here we fix this using the approach suggested in https://github.com/aiidalab/aiidalab/issues/196#issuecomment-899196469 Moreover, this seems to fix the segfault that we started seeing with `requests_cache>=1.0.0` described in #446.

Fixes #446 ( :crossed_fingers: ). Fixes #196 
Hopefully also fixes #413, I've removed the call to the app registry from tests.
 
### Testing 

Because we generally want to avoid fetching app index over network in tests, I've tested manually.

1. Launch a container with the latest `ghcr.io/aiidalab/full-stack:edge` image
2. Navigate to the app store page and observe that it crashes with "Kernel died" error, this is cause by the segfault described in #446 
3. Setup a dev environment with the help of aiidalab-dev
```
pip install git+https://github.com/aiidalab/aiidalab-dev
develop-aiidalab setup
```
4. Checkout this branch
```
cd local/aiidalab && git fetch origin fix-segfault && git switch fix-segfault
```
5. Now refresh the appstore page, it should now load.

Here's a script where you can reproduce the segfault from command line

```python
import logging
from home.app_store import AiidaLabAppStore    
      
logging.basicConfig(level='INFO')    
logging.getLogger('requests_cache').setLevel('DEBUG')    
      
store = AiidaLabAppStore()
```

TODO: Might be able to write a unit test with local file with either:
- https://pypi.org/project/requests-file/
- https://requests-mock.readthedocs.io/en/latest/index.html
- https://github.com/kiwicom/pytest-recording

### Debugging

Add logging calls to `aiidalab.utils.load_app_registry_index` or `load_app_registry_entry`:
```python
        response = _session.get(f"{AIIDALAB_REGISTRY}/apps_index.json")
        import time
        logger.info(f"Time: {time.ctime(int(time.time()))} Used cache: {response.from_cache}")
        return response.json()
```

PR #448, opened 2024-08-12
Memoize _AiidaLabApp.is_registered()

danielhollas:
Currently, whether an App is registered in the App Store or not is determined via `_AiidaLabAppis_registered()` method, which **always** fetches the app index to determine this. Even though we use `requests_cache` to cache the HTTP request, this is still wasteful (especially in App Store where this method is called for all the apps). I think it is a fair assumption that the registered status will not change during the lifetime of the Python process, so we can determine it once and store it as a class attribute. This is what is done in this PR.

Moreover, in the specific case of App Store page, the whole index is fetched first and then we create `_AiidaLabApp` instances for all the apps in the index. In this case, we already know that the apps are registered, so we can simply pass this info via class constructor. (This change will need to be implemented in `aiidalab-home` after this is merged and released).

Follow up to #447, where the fact that we've been fetching the app index so many times was triggering a bug in `requests_cache`.

TODO: Update existing tests.

### DIFF
--- a/aiidalab/app.py
+++ b/aiidalab/app.py
@@ -151,6 +151,10 @@ class _AiidaLabApp:
             local_registry_entry = cls._registry_entry_from_path(app_path)
             remote_registry_entry = load_app_registry_entry(app_id)
             registry_entry = remote_registry_entry or local_registry_entry
+            # If the caller did not tell us whether the app was registered or not,
+            # we can determine that here.
+            if registered is None:
+                registered = True if remote_registry_entry else False
 
         return cls.from_registry_entry(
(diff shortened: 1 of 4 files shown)


aiidalab/app.py
```diff
@@ -131,6 +136,7 @@ def from_id(
         app_id: str,
         registry_entry: dict[str, Any] | None = None,
         apps_path: str | None = None,
+        registered: bool | None = None,
```
danielhollas, line 139: We should probably use the functools.cached_property decorator.

https://docs.python.org/3/library/functools.html#functools.cached_property

Issue #442, opened 2024-07-21
Error when parsing requirements


danielhollas:
The App Store crashes on aiidalab/full-stack:edge, with error:

```python
Traceback (most recent call last):
  File "/opt/conda/lib/python3.9/site-packages/packaging/requirements.py", line 102, in __init__
    req = REQUIREMENT.parseString(requirement_string)
  File "/opt/conda/lib/python3.9/site-packages/pyparsing/core.py", line 1141, in parse_string
    raise exc.with_traceback(None)
pyparsing.exceptions.ParseException: Expected string_end, found '#'  (at char 26), (line:1, col:27)

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/opt/conda/lib/python3.9/threading.py", line 980, in _bootstrap_inner
    self.run()
  File "/opt/conda/lib/python3.9/threading.py", line 917, in run
    self._target(*self._args, **self._kwargs)
  File "/opt/conda/lib/python3.9/site-packages/aiidalab/utils.py", line 147, in wrapped
    return func(instance, *args, **kwargs)
  File "/opt/conda/lib/python3.9/site-packages/aiidalab/app.py", line 931, in refresh
    self._refresh_versions()
  File "/opt/conda/lib/python3.9/site-packages/aiidalab/app.py", line 912, in _refresh_versions
    all_available_versions = list(self._app.available_versions(prereleases=True))
  File "/opt/conda/lib/python3.9/site-packages/aiidalab/app.py", line 184, in available_versions
    version_requirements = [
  File "/opt/conda/lib/python3.9/site-packages/aiidalab/app.py", line 185, in <listcomp>
    Requirement(r)
  File "/opt/conda/lib/python3.9/site-packages/packaging/requirements.py", line 104, in __init__
    raise InvalidRequirement(
packaging.requirements.InvalidRequirement: Parse error at "'# Hotfix'": Expected string_end
```

danielhollas:
I think this is a duplicate of #326

Issue #326, opened 2022-11-24
Format error in requirement of app can maliciously break AppStore 


unkcpz:
The version compatibility check will check the format of the dependencies text, and the unscathed exception raise and break the AppStore if the requirements parse fails. The problem raised with new QeApp since I put the `numpy~=1.23 # pined for pymatgen` where the `#` will be regard as invalid symbol. The appstore raise:

```
---------------------------------------------------------------------------
ParseException                            Traceback (most recent call last)
/opt/conda/lib/python3.9/site-packages/packaging/requirements.py in __init__(self, requirement_string)
    112         try:
--> 113             req = REQUIREMENT.parseString(requirement_string)
    114         except ParseException as e:

/opt/conda/lib/python3.9/site-packages/pyparsing.py in parseString(self, instring, parseAll)
   1954                     exc.__traceback__ = self._trim_traceback(exc.__traceback__)
-> 1955                 raise exc
   1956         else:

/opt/conda/lib/python3.9/site-packages/pyparsing.py in parseImpl(self, instring, loc, doActions)
   3813         if loc < len(instring):
-> 3814             raise ParseException(instring, loc, self.errmsg, self)
   3815         elif loc == len(instring):

ParseException: Expected stringEnd, found '#'  (at char 12), (line:1, col:13)

During handling of the above exception, another exception occurred:

InvalidRequirement                        Traceback (most recent call last)
/opt/aiidalab-home/home/app_manager.py in _uninstall_app(self, _)
    453         """Attempt to uninstall the app."""
    454         try:
--> 455             self._check_detached_state()
    456             self.app.uninstall_app()
    457         except RuntimeError as error:

/opt/aiidalab-home/home/app_manager.py in _check_detached_state(self)
    415     def _check_detached_state(self):
    416         """Check whether the app is in a detached state which would prevent any install or other operations."""
--> 417         self.app.refresh()
    418         self._refresh_widget_state()
    419         blocked = self.app.detached and not self.blocked_ignore.value

/opt/conda/lib/python3.9/site-packages/aiidalab/utils.py in wrapped(instance, *args, **kwargs)
    140 
    141             self.last_start[hash(instance)] = time.perf_counter()
--> 142             return func(instance, *args, **kwargs)
    143 
    144         return wrapped

/opt/conda/lib/python3.9/site-packages/aiidalab/app.py in refresh(self)
    739                 self._refresh_versions()
    740                 self.set_trait(
[...]

PR #449, opened 2024-08-12
Skip invalid python requirements

danielhollas:
Fixes https://github.com/aiidalab/aiidalab/issues/442. Fixes https://github.com/aiidalab/aiidalab/issues/326

Stacked on top of #447

### DIFF
--- a/tests/static/app_registry.yaml
+++ b/tests/static/app_registry.yaml
@@ -15,7 +15,7 @@ releases:
     v22.11.0:
         environment:
             python_requirements:
-                - Jinja2~=2.11.3
+                - 'Jinja2~=2.11.3 # invalid requirement'
                 - aiida-core~=1.0
                 - aiida-quantumespresso~=3.5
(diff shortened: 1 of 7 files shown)


tests/static/app_registry.yaml
```diff
@@ -15,7 +15,7 @@ releases:
     v22.11.0:
         environment:
             python_requirements:
-                - Jinja2~=2.11.3
+                - 'Jinja2~=2.11.3 # invalid requirement'
```
superstar54, line 18: Does the test check this?

superstar54 (changes requested):
Hi @danielhollas , thanks for the work. I have a question regarding the test.

tests/static/app_registry.yaml
danielhollas: Thanks, good point. The test checked this implicitly since previously this would cause an exception. I've added an explicit unit test in `test_appdataclass.py`.

danielhollas:
I've added a docstring. I have also inlined all imports from `packaging.requirements` to places where there are actually used, to hopefully minimize a chance that a similar bug is introduced in the future.

superstar54 (approved):
LGTM! thanks for the nice work!

PR #450, opened 2024-08-12
Add more types to aiidalab.app

danielhollas:
Part of #363

### DIFF
--- a/aiidalab/app.py
+++ b/aiidalab/app.py
@@ -447,7 +447,7 @@ class _AiidaLabApp:
             with tempfile.TemporaryDirectory() as tmp_dir:
                 with tarfile.open(path) as tar_file:
 
-                    def is_within_directory(directory: Any, target: Any) -> Any:
+                    def is_within_directory(directory: str, target: str) -> bool:
                         abs_directory = os.path.abspath(directory)
                         abs_target = os.path.abspath(target)
 
@@ -455,7 +455,7 @@ class _AiidaLabApp:
 
                         return prefix == abs_directory
 
-                    def safe_extract(tar: Any, path: Any) -> None:
+                    def safe_extract(tar: tarfile.TarFile, path: str) -> None:
                         for member in tar.getmembers():
                             member_path = os.path.join(path, member.name)
                             if not is_within_directory(path, member_path):
@@ -764,7 +764,7 @@ class AiidaLabApp(traitlets.HasTraits):  # type: ignore
         else:
             self._watch = None  # type: ignore
 
-    def __str__(self):
+    def __str__(self) -> str:
         return f"<AiidaLabApp name='{self._app.name}'>"
 
     @traitlets.default("include_prereleases")
--- a/aiidalab/app.py
+++ b/aiidalab/app.py
@@ -863,7 +863,7 @@ class AiidaLabApp(traitlets.HasTraits):  # type: ignore
             self.refresh()
             return version
 
-    def uninstall_app(self, _: Any = None) -> None:
+    def uninstall_app(self) -> None:
         """Uninstall application."""
         # Perform uninstall process.
         with self._show_busy():
--- a/aiidalab/app.py
+++ b/aiidalab/app.py
@@ -205,9 +205,11 @@ class _AiidaLabApp:
                 ) and self._strict_dependencies_met(version_requirements, python_bin):
                     yield version
 
-    def dirty(self):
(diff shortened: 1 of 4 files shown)


aiidalab/app.py
```diff
@@ -484,13 +486,13 @@ def _install_from_git_repository(self, git_url):
...
-        python_bin=None,
-        install_dependencies=True,
-        stdout=sys.stdout,
-        prereleases=False,
-        post_install_triggers=True,
-    ):
+        version: str | None = None,
+        python_bin: str | None = None,
+        install_dependencies: bool = True,
+        stdout: Any = sys.stdout,
```
danielhollas, line 492: `sys.stdout` has type `Union[typing.TextIO, Any]` so for conciseness we'll simply use Any.

setup.cfg
```diff
@@ -73,17 +73,11 @@ docs =
...
 
 [mypy]
+strict = True
 ignore_missing_imports = True
 strict_optional = True
 show_error_codes = True
 
-# strictness
-strict = True
-# these options reduce the strictness and should eventually be removed
```
danielhollas, line 82: This is an outdated comment since all these options have been set to True. At this point they are superfluous since they are implied by `strict=True`.

aiidalab/app.py
```diff
@@ -861,8 +863,8 @@ def update_app(
             self.refresh()
             return version
 
-    def uninstall_app(self, _=None):
-        """Perfrom app uninstall."""
+    def uninstall_app(self, _: Any = None) -> None:
+        """Uninstall application."""
```
superstar54, line 867: This does not related to this PR, but why do we add a parameter which is not used?

aiidalab/app.py
```diff
@@ -674,7 +676,7 @@ def stop(self) -> None:
         if self._monitor_thread is not None:
             self._monitor_thread_stop.set()
 
-    def is_alive(self) -> bool | None:
+    def is_alive(self) -> bool | None | Thread:
```
superstar54, line 679: I don't think it will return a `Thread` or `None`.

superstar54 (changes requested):
Hi @danielhollas , thanks for the work. I added some small change requests.

aiidalab/app.py
danielhollas: Good point, the only instance where I found this function used is in `aiidalab-home`, and it doesn't pass any arguments. 

https://github.com/aiidalab/aiidalab-home/blob/2b6ea6d68bcd2edcbb6fb8689338fc721d06c360/home/app_manager.py#L510

It's possibly a historical artifact when this function was called as part of the traitlets / ipywidgets callback. I'll remove it.

aiidalab/app.py
danielhollas: `mypy` says otherwise. :sweat_smile: 

```
aiidalab/app.py:681: error: Incompatible return value type
 (got "Thread | bool | None", expected "bool | None")  [return-value]
```

I was extremely surprised by this, but it turns out that in Python `and` and `or` operators do not return bool values, they return one of the operands. See the [reference documentation](https://docs.python.org/3/library/stdtypes.html#boolean-operations-and-or-not) and https://stackoverflow.com/a/28321263/3682277.

superstar54: 😅. In this case, it will never return `Thread`. But we can keep it if `mypy` complains.

superstar54 (approved):
LGTM!

aiidalab/app.py
danielhollas: You're right, that is weird. I think this is an instance of this issue: https://github.com/python/mypy/issues/16565

In principle, the Thread instance should always be truthy and therefore never returned, *unless* we overwrote its `__bool__` method. I think that's why mypy can't correctly infer this in this case. More info in official docs: https://docs.python.org/3/library/stdtypes.html#truth-value-testing

> By default, an object is considered true unless its class defines either a [__bool__()](https://docs.python.org/3/reference/datamodel.html#object.__bool__) method that returns False or a [__len__()](https://docs.python.org/3/reference/datamodel.html#object.__len__) method that returns zero, when called with the object. [[1]](https://docs.python.org/3/library/stdtypes.html#id12) Here are most of the built-in objects considered false:

Here's a minimal reproducer on mypy playground.
https://mypy-play.net/?mypy=latest&python=3.12&gist=ec46f8172f3e4b8ea65def242d9bc3c4

PR #453, opened 2024-08-21
Finish typing of aiidalab.app module

danielhollas:


aiidalab/app.py
```diff
@@ -585,7 +591,7 @@ class AppPathFileSystemEventHandler(FileSystemEventHandler):  # type: ignore
         def __init__(self, app: AiidaLabApp):
             self.app = app
 
-        def on_any_event(self, event):
+        def on_any_event(self, event: FileSystemEvent) -> None:
```
danielhollas, line 594: I looked inside the `watchdog` package to see the correct type here.

.pre-commit-config.yaml
```diff
@@ -45,6 +45,7 @@ repos:
                 - types-requests
                 - types-toml
                 - types-click
+                - traitlets~=5.0
```
danielhollas, line 48: pre-commit runs mypy from an isolated virtual environment so it doesn't have access to the installed dependencies, and thus we need to add them here manually. This is highly problematic, as we are essentially duplicating info from setup.cfg. In the future, we probably have to switch to an approach used in aiida-core, where mypy is installed as dev dependency of aiidalab, instead of being installed by pre-commit.

aiidalab/app.py
```diff
@@ -914,9 +924,12 @@ def _refresh_versions(self) -> None:
         )
 
     def _refresh_dependencies_to_install(self) -> None:
-        self.dependencies_to_install = self._app.find_dependencies_to_install(
-            self.version_to_install
-        )
+        if self.version_to_install:
```
danielhollas, line 927: This could be None so mypy got rightfully mad that we're passing a possible None value to a function that did not expect it. In other words, this was a bug.

aiidalab/app.py
```diff
@@ -28,7 +28,7 @@
 from dulwich.errors import NotGitRepository
 from packaging.requirements import Requirement
 from packaging.version import parse
-from watchdog.events import EVENT_TYPE_OPENED, FileSystemEventHandler
+from watchdog.events import EVENT_TYPE_OPENED, FileSystemEvent, FileSystemEventHandler
```
BoboTiG, line 31: Nitpick: the new import could be moved under the `if TYPE_CHECKING:` condition.

BoboTiG:
I released watchdog [v5.0.0](https://github.com/gorakhargosh/watchdog/releases/tag/v5.0.0) that contains a full types coverage, it should ease your job here :)

And, again, thank you for testing the `madter` branch at the time, it helped catch a bunch of problems.

danielhollas:
@superstar54 will you have time to look at this? I'd like to include this in the next release this week.

aiidalab/__main__.py
```diff
@@ -11,19 +11,22 @@
...
 from textwrap import indent, wrap
+from typing import TYPE_CHECKING, Any, Generator
 
 import click
 from click_spinner import spinner
 from tabulate import tabulate
 
 from . import __version__
-from .app import AppVersion
-from .app import _AiidaLabApp as AiidaLabApp
```
danielhollas, line 21: This renaming was super confusing.

aiidalab/__main__.py
```diff
@@ -560,7 +585,7 @@ def uninstall(app_name, yes, dry_run, force, fully_remove):
 
 
 @contextmanager
-def _mock_schemas_endpoints():
+def _mock_schemas_endpoints() -> Generator[None, None, None]:
```
unkcpz, line 588:
```suggestion
def _mock_schemas_endpoints() -> Iterator[None]:
```

I think this is simpler?

aiidalab/app.py
```diff
@@ -832,12 +843,12 @@ def _validate_version_to_install(self, proposal):
...
 
     @traitlets.observe("version_to_install")
-    def _observe_version_to_install(self, change):
+    def _observe_version_to_install(self, change: dict[str, Any]) -> None:
         if change["old"] != change["new"]:
             self.refresh()
 
     @contextmanager
-    def _show_busy(self):
+    def _show_busy(self) -> Generator[None, None, None]:
```
unkcpz, line 851:
```suggestion
    def _show_busy(self) -> Iterator[None]:
```

unkcpz (commented):
Thanks @danielhollas, just a minor request, all good then.

aiidalab/__main__.py
danielhollas: Hmm, I guess you're right, and mypy accepts this. However, since we use the Generator type in many other places now, I'd like to keep things consistent for now. Later if we like we can go through and simplify this accross the code base.

unkcpz: sure!

unkcpz (approved):

PR #454, opened 2024-08-28
Update watchdog dependency 

danielhollas:
### Context 

We use `watchdog` to monitor file events in app's directories so that we can potentially modify app's status immediately. (This is very cool although I think this is perhaps overengineered given a relatively niche use-case). See screencast below for a demonstration

### Motivation

Starting with version v5 watchdog library is now fully typed, which will help us towards #363. was released recently.

[Changelog for v5](https://github.com/gorakhargosh/watchdog/releases/tag/v5.0.0). 
[Changelog for v6](https://github.com/gorakhargosh/watchdog/releases/tag/v6.0.0). 

There is one change in the code to ignore the newly added `EVENT_TYPE_CLOSED_NO_WRITE` event which we do not care about

Citing from the inotify(7) manpage,

>   IN_CLOSE_NOWRITE is fired when a file
      that was opened in read-only mode was closed.

Since we only watch for events that (potentially) modify files inside the app git repo, we can ignore this event.

### Test plan

See screencast. NOTE: I've verified that the double-refresh is an pre-existing issue that's unrelated to this PR.

[Screencast from 2026-04-10 10-12-42.webm](https://github.com/user-attachments/assets/bf246cdc-02fd-49ed-a560-15d4eb7f13f5)

### DIFF
--- a/aiidalab/app.py
+++ b/aiidalab/app.py
@@ -27,7 +27,11 @@ from uuid import uuid4
 import requests
 import traitlets
 from dulwich.errors import NotGitRepository
-from watchdog.events import EVENT_TYPE_OPENED, FileSystemEventHandler
+from watchdog.events import (
+    EVENT_TYPE_CLOSED_NO_WRITE,
+    EVENT_TYPE_OPENED,
+    FileSystemEventHandler,
+)
 from watchdog.observers import Observer
 from watchdog.observers.polling import PollingObserver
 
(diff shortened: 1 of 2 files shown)


danielhollas:
I want to give it a manual test before merging, but otherwise should be ready. 
I'll merge *after* release of the new version.

danielhollas:
I've finally tested this and updated PR description with a screencast.
I'd like to get this in for a next release.

edan-bainglass (approved):
Very cool. LGTM! Might be useful to include a simple use case in the description. How do you see this being put to use in practice?

danielhollas:
Thanks. I don't know what the original use case for this feature was, that's a question for @yakutovicha . Here I am just bumped ng the dependency. As I mention above, it seemed a bit overengoneered and we might decide to remove as it might not be worth the maintanence cost.

PR #460, opened 2025-03-05
Policy templates

edan-bainglass:
This PR adds templates for Terms of Use and Privacy policies, as well as utilities for generating these documents for an AiiDAlab deployment. Dependencies to run the script are provided in a separate python reqs file.

### DIFF
--- a/docs/source/admin/index.rst
+++ b/docs/source/admin/index.rst
@@ -27,6 +27,17 @@ We provide `instructions and deployment k8s scripts <https://github.com/aiidalab
 
 .. _admin-guide:maintain-app-registry:
 
+
+Policy documents
+----------------
+
+When deploying AiiDAlab for multiple users, you should provide a privacy policy and terms of use.
+These legal documents should be accessible to users before they start using the service.
+We provide templates in the ``templates`` directory of the AiiDAlab repository.
+The templates were developed for Swiss-based deployments. Please adjust them to your needs.
+Utilities are provided in the ``templates`` directory to help you generate the documents.
+Please see the ``README.md`` file in the ``templates`` directory for more information.
+
 Maintaining an app registry
 ===========================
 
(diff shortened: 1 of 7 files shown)


edan-bainglass:
@eimrek maybe you can have a look, since you've at least touched the present concrete version on the website? 🙏

edan-bainglass:
@yakutovicha could really use a quick review on this one 🙏 These are all new files, so it shouldn't collide with anything. And the dependencies required for the local utility script is added in a reqs file, so not touching the main package. It is really only here due to its general purpose.

eimrek (commented):
Looks fine from my POV. But two possible comments (considering that I don't know this repo, so i'm not certain if they make sense):

* potentially  links to these documents could be added to the default start/login page of the deployment?
* and, the html files could automatically be generated from the template when starting the server, without requiring a manual step?

but it really depends on how deployment works in practice.

eimrek:
Also, try to copy the templates to an LLM. Mine detected places where you forgot to change "PSI" and "Switzerland" to the jinja variable. and additionally found some grammar / clarify problems.

edan-bainglass:
> Looks fine from my POV. But two possible comments (considering that I don't know this repo, so i'm not certain if they make sense):
> 
> * potentially  links to these documents could be added to the default start/login page of the deployment?
> * and, the html files could automatically be generated from the template when starting the server, without requiring a manual step?
> 
> but it really depends on how deployment works in practice. 

Interesting ideas for future consideration 🙂

edan-bainglass:
> Also, try to copy the templates to an LLM. Mine detected places where you forgot to change "PSI" and "Switzerland" to the jinja variable. and additionally found some grammar / clarify problems.

Found one instance of PSI - replaced with `{{ institution.name.short }}`. As for Switzerland, we discussed internally to leave this in the template and clearly state that the templates were prepared for Swiss use. External users should take extra care (in consultation with the relevant department) in preparing their documents.

edan-bainglass:
Moving this to the [demo server repo](https://github.com/aiidalab/aiidalab-demo-server)

PR #461, opened 2025-03-26
Update documentation

edan-bainglass:
This PR does the following:
1. Subs `sphinx-panels` for `sphinx-design`
2. Updates Usage pages, specifically documentation regarding local Docker-based installation


edan-bainglass:
@superstar54 the idea is to have the general AiiDAlab installation here, then reduce the text on AiiDAlab QE docs by simply referencing this documentation with a comment regarding the different Docker image.

One thing to note is that in the QE docs, the VM install approach is more verbose, not just a link to the Quantum Mobile download page. I can't remember why we did this, but if desired, I can add this page here.

edan-bainglass:
> @superstar54 the idea is to have the general AiiDAlab installation here, then reduce the text on AiiDAlab QE docs by simply referencing this documentation with a comment regarding the different Docker image.
> 
> One thing to note is that in the QE docs, the VM install approach is more verbose, not just a link to the Quantum Mobile download page. I can't remember why we did this, but if desired, I can add this page here.

Actually, Giovanni stated in some recent email that we no longer recommend the VM route. @giovannipizzi should we remove the VM option from the docs then?

edan-bainglass:
Okay, maybe for now, we don't touch these two. We can update it later, particularly after I update aiidalab.net/deployments

superstar54 (changes requested):
Thanks for the work, moving the part from QEapp to here is better.
I had a quick look. Here are a few suggestions:

- In the Local Docker Instance section, the recommended way to launch AiiDAlab depends on the system. On a Windows system, the Docker Desktop is recommended. Many experimentists don't know WSL or the command line, so the Docker Desktop is preferred over aiidalab-launch.
- The AiiDAlab Home page is too short, suggest adding one screenshot of the home page.
- suggest merging the ``App Store " section with the " Managing apps` section.

edan-bainglass:
> Thanks for the work, moving the part from QEapp to here is better. I had a quick look. Here are a few suggestions:
> 
> * In the Local Docker Instance section, the recommended way to launch AiiDAlab depends on the system. On a Windows system, the Docker Desktop is recommended. Many experimentists don't know WSL or the command line, so the Docker Desktop is preferred over aiidalab-launch.

Valid. Will try to adjust for this.

> * The AiiDAlab Home page is too short, suggest adding one screenshot of the home page.

Agreed

> * suggest merging the ``App Store " section with the " Managing apps` section.

I prefer it separate. Even though one can get to app management from the store page, we are showing screenshots for doing so from the home page. So in my opinion, best to keep these separate.

docs/source/usage/access/launch.rst
```diff
@@ -0,0 +1,65 @@
...
+It helps to manage multiple AiiDAlab profiles, each with its own home directory for persistent storage, and allows to easily switch between them.
+To use AiiDAlab launch, make sure to first **install Python** (if not already available), then
+
+#. Install AiiDAlab launch with `pipx <https://pypa.github.io/pipx/installation/>`_ (**recommended**):
+
+   .. code-block:: console
+
+      pipx install aiidalab-launch
+
+   or directly with pip
```
mikibonacci, line 20: I had some issue with the pipx installed version. not sure why, but when I installed via pip, I was then able to start aiidalab without issues. Could be a problem of old packages, which maybe pipx does not manage properly. Anyway, I am reporting this as maybe we can just add a short sentence: "if you installed via pipx but you have some issue in aiidalab-launch start, try to install it via pip" ... or something like that.

mikibonacci (changes requested):
@edan-bainglass , LGTM, just a small request on aiidalab-launch page

docs/source/usage/access/launch.rst
edan-bainglass: Thanks @mikibonacci. I've never tried the `pipx` approach. I'll open an issue to remind us to check for compatibility. For now, I'll add a comment as you suggest.

mikibonacci: for reference, this is the error: `docker.errors.DockerException: Error while fetching server API version: Not supported URL scheme http+docker`

edan-bainglass: Pinging @yakutovicha, the OG developer, assuming he might be familiar with the `pipx` approach.

edan-bainglass: Perhaps we should remove the recommendation? Strange to recommend `pipx` but also add a comment that it may not work. What do you think?

edan-bainglass: Thanks for your comment @danielhollas. I will add the comment then, as @mikibonacci suggested.

mbercx:
> @mbercx can you please comment on this? 🙏

The last time I worked on the Quantum Mobile - mostly on getting the UTM version for MacOS - I hadn't really considered AiiDAlab yet. I suppose the question is if we'd consider it valuable to have this option? Do we think a lot of users will want to use a virtual machine over e.g. Docker?

edan-bainglass:
I tried the latest Quantum Mobile image in Virtual box and got the following when trying to start the machine:

image

Any ideas?

edan-bainglass:
To avoid holding this off (which is holding QE doc updates off), I will proceed with merging. We can address the matter of the Quantum Mobile option in a follow-up PR.

mikibonacci (approved):
Hi @edan-bainglass LGTM!

edan-bainglass:
@superstar54 updated the relevant sections per our discussion. Please review 🙏

superstar54 (changes requested):
Could you add some screenshots for the Windows installation?

edan-bainglass:
> Could you add some screenshots for the Windows installation?

My bad. Didn't push my changes. Check now 🙏

superstar54:
Thanks, very nice! 

One thing I'm unsure about is the `.conda` folder. I previously suggested removing this step, but I might have been wrong. I noticed that in the `aiidalab-launch`, a volume is also created for the `.conda` folder. So, please **add this back** — no need to include a screenshot. Just mention that an additional volume should be added for `.conda`, similar to the one for `home`. After this, I will approve the PR. Apologies for the back and forth!

edan-bainglass:
> add this back

To be clear, this is why I had it there in the first place 😅, to mimic aiidalab-launch. But is it really necessary, now that we no longer use it as you stated?

superstar54:
> But is it really necessary, now that we no longer use it as you stated?

I was wrong; it might be used to preserve the user's installation environment. Let's keep it untouched for the moment, so please add this back.

@danielhollas, could you comment on the usage of the `.conda` folder?

edan-bainglass:
@superstar54 done!

PR #462, opened 2025-04-02
Add documentation section on deploying usage policies

edan-bainglass:
Policy documents are now templated and deployed by each deployment (no longer housed here). This PR adds a note in the documentation regarding the use of these templates.


superstar54 (approved):
LGTM!

PR #464, opened 2025-04-03
Fix: Update Code to Pass `mypy` Checks

ayushjariyal:
issue #363 ,

While running `mypy` to some files, they failed. So i update them to pass all `mypy` checks.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -36,22 +36,16 @@ repos:
       hooks:
           - id: check-github-workflows
 
-    - repo: https://github.com/pre-commit/mirrors-mypy
-      rev: v1.14.0
+    - repo: local
       hooks:
           - id: mypy
+            name: mypy
+            entry: mypy
             args: [--config-file=setup.cfg]
-            additional_dependencies:
-                - types-tabulate
-                - types-setuptools
-                - types-jsonschema
-                - types-cachetools
-                - types-requests
-                - types-toml
-                - types-click
-                - types-click-spinner
-                - traitlets~=5.0
-                - packaging
+            language: python
+            types: [python]
+            require_serial: true
+            pass_filenames: true
             exclude: >-
                 (?x)^(
                   docs/.*|
--- a/setup.cfg
+++ b/setup.cfg
@@ -48,6 +48,15 @@ dev =
     pytest~=8.2.0
     pytest-cov~=5.0
     ruamel.yaml~=0.16
+    mypy~=1.14.0
+    types-tabulate
+    types-setuptools
+    types-jsonschema
+    types-cachetools
+    types-requests
+    types-toml
+    types-click
+    types-click-spinner
 registry =
     CacheControl~=0.12
     jsonref~=0.2
@@ -77,6 +86,10 @@ strict = True
 ignore_missing_imports = True
 strict_optional = True
 show_error_codes = True
+exclude = (?x)(
+    *tests/*$
+    | docs/*$
+  )
 
 [mypy-aiidalab.registry.*]
 ignore_errors = True


danielhollas:
Thanks!

> While running mypy to some files, they failed. So i update them to pass all mypy checks.

It seems that mypy fails on this PR in the pre-commit.ci job, see 
https://results.pre-commit.ci/run/github/136350008/1743686672.PFMzNdfrTDeMdI1WrFhg3A

Out of curiosity, which Python version are you using to run mypy locally?

Also, please revert the version change in `setup.cfg`  and `aiidalab/__init__.py`.

ayushjariyal:
@danielhollas , I was using Python 3.9.20, but now I have updated it to Python 3.13.2.

ayushjariyal:
Hey @danielhollas ,I'm not sure why my `pre-commit` checks are failing, even though `mypy` passes all checks when I run it locally. I'm using `Python 3.13.2` and `mypy 1.13.0`.

danielhollas:
@ayushjariyal you're right. The way that we run mypy through pre-commit is not correct since it doesn't install the package dependencies. We'll need to fix that before this can be merged. We basically will want to run mypy in the same way as in aiida-core repo: https://github.com/aiidateam/aiida-core/blob/660fec70ef43a64be7edae3c12f0a0fd5ef84349/.pre-commit-config.yaml#L78

But that also involved adding mypy and the various `types-*` packages to dev dependencies in `setup.cfg`.
Feel free to give it a shot if you want, otherwise I'll do it at some point, but not in the next two weeks since I am away on a conference and on holidays.

Thanks for the PR!

ayushjariyal:
Thank you for the response! I'll give it a shot and try to fix it.

danielhollas:
Hmm, turns out, when you do the change, we will have to skip the mypy check in `pre-commit.ci` since it doesn't support installing the dependencies from setup.cfg. That can be done at the top of  `.pre-commit-config.yaml` via the `skip` key, see: 

https://github.com/aiidateam/aiida-core/blob/main/.pre-commit-config.yaml#L5

Then we'll have to add a new pre-commit job to Github Actions, where we will run pre-commit manually, see here for example: 

https://github.com/aiidateam/aiida-core/blob/main/.github/workflows/release.yml#L31

It's all a bit tricky so feel free to ask questions if you get stuck somewhere.

.github/workflows/publish.yml
```diff
@@ -43,6 +43,29 @@ jobs:
                   name: release
                   path: dist/
 
+    pre-commit:
```
yakutovicha, line 46: Why would this be necessary? The pre-commit test is run by [pre-commit.ci](https://pre-commit.ci/), so that seems like a redundant thing?

danielhollas: It's needed, because to run mypy correctly we need to install all the dependencies, and `pre-commit.ci` doesn't allow that. We have the same problem in aiida-core, where we recently removed this "redundant" pre-commit job without understanding the ramifications. I'll open an issue there with more details.

danielhollas, line 46: @ayushjariyal thanks for adding this, but please move it to ci.yml so that it runs on every PR. Also please use uv to install aiidalab package dependencies. You don't need to install pre-commit explicitly as it should be part of the `dev` extras. See how we do it here:

https://github.com/aiidalab/aiidalab/blob/main/.github%2Fworkflows%2Fci.yml#L34

Also if you haven't done that already, install pre-commit locally and then run:

```console
pre-commit install
```

That way the pre-commit hooks will run automatically whenever you make a new commit.

danielhollas (changes requested):

ayushjariyal:
Thank you for the update! No problem at all—I'll wait for your ping.

PR #469, opened 2025-04-09
Update mypy and fix its configuration

### LINKED ISSUE #363: Complete the mypy check for whole package

### LINKED ISSUE #459: Resolve mypy v1.14 failures
Downgraded mypy for now since v1.14 fails with 

```
aiidalab/metadata.py:61: error: Argument 2 to "get" of "SectionProxy" has incompatible type "str | Any | None"; expected "str | list[str]"  [arg-type]
aiidalab/metadata.py:62: error: Argument 2 to "get" of "SectionProxy" has incompatible type "str | Any | None"; expected "str | list[str]"  [arg-type]
aiidalab/metadata.py:63: error: Argument 2 to "get" of "SectionProxy" has incompatible type "str | Any | None"; expected "str | list[str]"  [arg-type]
aiidalab/metadata.py:64: error: Argument 2 to "get" of "SectionProxy" has incompatible type "str | Any | None"; expected "str | list[str]"  [arg-type]
aiidalab/metadata.py:65: error: Argument 2 to "get" of "SectionProxy" has incompatible type "str | Any | None"; expected "str | list[str]"  [arg-type]
aiidalab/metadata.py:72: error: Argument 2 to "get" of "SectionProxy" has incompatible type "str | None"; expected "str | list[str]"  [arg-type]
aiidalab/metadata.py:77: error: Argument 2 to "get" of "SectionProxy" has incompatible type "str | None"; expected "str | list[str]"  [arg-type]
```

_Originally posted by @danielhollas in https://github.com/aiidalab/aiidalab/issues/458#issuecomment-2616088004_

### LINKED ISSUE #477: Fix mypy pre-commit config
https://github.com/python/mypy/issues/13916

danielhollas:
The main change here is that we stop using the mypy hook (mirrors-mypy) and instead run mypy from within the same virtual environment with all other dependencies, which is needed for more accurate type checking. This is the same configuration that is used in aiida-core. See also discussion about this well known issue on mypy's own issue tracker: https://github.com/python/mypy/issues/13916

Because of the limitation of `pre-commit.ci` service, which prohibits installing dependencies outside of pre-commit hooks, we need to add a new job workflow in GHA. Again, this is the same as in aiida-core, see https://github.com/aiidateam/aiida-core/issues/6831

Split out from #464.

Closes #459. Closes #363. Closes #477


aiidalab/git_util.py
```diff
@@ -92,12 +92,12 @@ def get_branch_tracking_status(self, branch: bytes) -> BranchTrackingStatus | No
                 return BranchTrackingStatus.EQUAL
 
             # Check if local branch is behind the tracked branch:
-            for commit in self.get_walker(self.refs[tracked_branch]):
+            for commit in self.get_walker([self.refs[tracked_branch]]):
```
danielhollas, line 95: These calls were working because of this code path in dulwich:

https://github.com/jelmer/dulwich/blob/3376d1f99806323305ee4666146367bdcda7d02e/dulwich/walk.py#L281

aiidalab/git_util.py
```diff
@@ -20,10 +20,10 @@
 from enum import Enum
 from pathlib import Path
 from subprocess import CalledProcessError, run
-from typing import Any
 from urllib.parse import urldefrag
 
 from dulwich.porcelain import branch_list, status
+from dulwich.refs import Ref  # Ref = NewType(bytes)
```
danielhollas, line 26: https://docs.python.org/3/library/typing.html#newtype

.pre-commit-config.yaml
```diff
@@ -1,6 +1,7 @@
 ---
 ci:
     autoupdate_schedule: quarterly
+    skip: [mypy]
```
danielhollas, line 4: In `pre-commit.ci` we can't install mypy from pyproject toml so we need to skip this check. But I've added a dedicated pre-commit job because of that.

.pre-commit-config.yaml
```diff
@@ -31,26 +32,20 @@ repos:
...
           - id: mypy
-            args: [--config-file=pyproject.toml]
-            additional_dependencies:
-                - types-tabulate
-                - types-setuptools
-                - types-jsonschema
-                - types-cachetools
-                - types-requests
-                - types-toml
-                - types-click
```
danielhollas, line 50: `types-click` is no longer needed since the click library itself is now typed.
The rest of these has been moved to `pyproject.toml`

.pre-commit-config.yaml
```diff
@@ -31,26 +32,20 @@ repos:
...
-                - types-click-spinner
-                - traitlets~=5.0
-                - packaging
+            name: mypy
+            entry: mypy
+            args: [--config-file=pyproject.toml, aiidalab/]
+            language: python
+            types: [python]
+            require_serial: true
+            pass_filenames: false
```
danielhollas, line 42: It's important to run `mypy` on all files, not just the changed files, since changes in one file can induce type errors in another file.

.pre-commit-config.yaml
```diff
@@ -31,26 +32,20 @@ repos:
...
-                - types-requests
-                - types-toml
-                - types-click
-                - types-click-spinner
-                - traitlets~=5.0
-                - packaging
+            name: mypy
+            entry: mypy
+            args: [--config-file=pyproject.toml, aiidalab/]
+            language: unsupported
```
danielhollas: https://pre-commit.com/#unsupported

.pre-commit-config.yaml
```diff
@@ -12,45 +13,33 @@ repos:
...
-      rev: v0.15.0
+      rev: v0.15.10
       hooks:
           - id: ruff-format
             exclude: ^docs/.*
           - id: ruff
             args: [--fix, --exit-non-zero-on-fix, --show-fixes]
 
-    #  Currently disabled due to issues with installation pre-commit.ci:
-    #- repo: https://github.com/mgedmin/check-manifest
```
danielhollas, line 23: We're no longer using MANIFEST file in this repo

edan-bainglass (approved):
LGTM! Honestly, this is a bit much to review in detail (too many questions). But if the aim (even if partially) is to standardize w.r.t aiida-core, then 👍

danielhollas:
> Honestly, this is a bit much to review in detail (too many questions).

For sure, there's a lot of subtle points, that would be kind of annoying to write down everything. :sweat_smile: Happy to hop on a quick call tomorrow afternoon in case you're interested in getting answers to you questions. :-) 

> But if the aim (even if partially) is to standardize w.r.t aiida-core, then 👍

The main goal here is to make the typing here stronger. The point of comparing to aiida-core is just to say that I am not making this all up and there is a precedent in doing things this way. :-)

PR #470, opened 2025-04-11
Improve local installation docs

edan-bainglass:


mikibonacci (approved):
LGTM

docs/source/usage/access/desktop.rst
```diff
@@ -22,9 +22,18 @@ Fetch the AiiDAlab image
...
-      * ``aiidalab/qe`` - pre-configured with the `Quantum ESPRESSO app <https://aiidalab-qe.readthedocs.io/index.html>`_ (**recommended**)
+      * ``aiidalab/qe`` - pre-configured with the Quantum ESPRESSO app (**recommended**)
       * ``aiidalab/full-stack`` - bare image
 
+        .. note::
+
+            .. raw:: html
+
+               <p>
+                  We recommend the pre-configured image, as it includes much of the mechanics necessary to run a calculation on AiiDAlab.
```
danielhollas, line 33: This is only true when the user plans to use QeApp so I think this sentence should say as much, given that these are docs for general usage, right?

docs/source/usage/access/launch.rst
```diff
@@ -7,41 +7,66 @@ AiiDAlab launch
 
    The following steps require a local installation of Docker. You can verify your Docker installation by running ``docker run hello-world`` in the terminal.
 
-`AiiDAlab launch`_ is a thin Docker wrapper which takes care of all the prerequisites to run the AiiDAlab Docker image.
-It helps to manage multiple AiiDAlab profiles, each with its own home directory for persistent storage, and allows to easily switch between them.
-To use AiiDAlab launch, make sure to first **install Python** (if not already available), then
+.. raw:: html
```
danielhollas, line 10: Is there a reason to writing raw HTML? Seems harder to maintain...

docs/source/usage/access/launch.rst
```diff
@@ -7,41 +7,66 @@ AiiDAlab launch
...
+      <a href="https://github.com/aiidalab/aiidalab-launch" target="_blank">AiiDAlab launch</a> is a thin Docker wrapper which takes care of all the prerequisites to run the AiiDAlab Docker image.
+      It helps to manage multiple AiiDAlab profiles, each with its own home directory for persistent storage, and allows to easily switch between them.
+      To use AiiDAlab launch, please follow these steps:
+   </p>
 
-   .. code-block:: console
+#. Install AiiDAlab launch using one of the following methods:
 
-      pipx install aiidalab-launch
+   * ``pipx`` (**recommended**)
```
danielhollas, line 20: Should we point users to uv instead of pipx? It's a much more general tool. Happy to push a change to this PR or do it in a follow-up.

docs/source/usage/access/launch.rst
```diff
@@ -7,41 +7,66 @@ AiiDAlab launch
...
 
       aiidalab-launch profile add --image aiidalab/qe:latest aiidalab
 
-   * Bare AiiDAlab
+     .. tip::
+
+         .. raw:: html
+
+            <p>
+               We recommend this pre-configured image, as it includes much of the mechanics necessary to run a calculation on AiiDAlab.
```
danielhollas, line 59: Same comment as above, if users don't plan on using QeApp, it's wasteful to download the QeApp image.

danielhollas (changes requested):
Thanks for working on the docs! 🙏

docs/source/usage/access/desktop.rst
edan-bainglass: Requested by Giovanni

docs/source/usage/access/launch.rst
edan-bainglass: To make the link external, so users don't lose our docs

docs/source/usage/access/launch.rst
edan-bainglass: Follow up PR

docs/source/usage/access/launch.rst
edan-bainglass: Same response

docs/source/usage/access/desktop.rst
danielhollas: To clarify, I am not asking to remove this, just to start the sentence with (`If you plan to use QeApp, ....` or something along those lines.)

edan-bainglass: Understood, but same effect. The recommendation requested by @giovannipizzi is not from an assumption that the user will use the QE app, but rather from a suggestion that we recommend an image with at least one app installed (particularly our "flagship" app), so that users can get to play around quicker. But how about I let @giovannipizzi comment here. I'm happy to modify if agreed upon.

giovannipizzi:
Thanks! All very good, only thing: it would be good to also have the same structure (panel + button) to go to the demo, so it's visually clear that it's also an option to run aiidalab as the other ones, and one knows very quickly where to click without reading and figuring out which is the correct link.

giovannipizzi:
Very nice!

Issue #472, opened 2025-04-18
Add `/venv*/` to `.gitignore` to avoid commiting virtual environment files


ayushjariyal:
Currently, the repository does not include virtual environment folder (e.g., `/venv*/`) in the `.gitignore` file. This means if someone creates a virtual environment inside the project folder, it might accidentally get added to Git when committing changes.

Virtual environment are usually specific to each person's machine and are not meant to be shared in the repository. To avoid accidentally uploading these files, we can add them to the `.gitignore` file.

ayushjariyal:
Hi @danielhollas,
I was going through the repository and notice that virtual environment folder is not included in the `.gitignore` file. I was wondering if it would be helpful to add this, so user don't accidentally commit their local virtual environments.

Would this be a relevant issue? If So, then I'd be happy to submit a PR for this.

danielhollas:
Closed in #473

PR #473, opened 2025-04-19
Adding venv to .gitignore file

ayushjariyal:
issue #472

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -7,10 +7,8 @@ dist/
 build/
 docs/_build/
 apps_meta.sqlite
-Pipfile.lock
-requirements.txt_dev
-requirements.txt_stable
 .DS_Store
 .coverage
 
-/venv*/
+.venv*/
+venv*/
--- a/.gitignore
+++ b/.gitignore
@@ -12,3 +12,5 @@ requirements.txt_dev
 requirements.txt_stable
 .DS_Store
 .coverage
+
+/venv*/


danielhollas:
Thanks for the PR @ayushjariyal!

Just out of curiosity, how do you create your virtual environment? And if you run `git status` do you see the venv folder as untracked?

In my case, I am using [uv](https://github.com/astral-sh/uv) and create environment with

```console
uv venv
```

which creates a folder `.venv/` containing a `CACHEDIR.TAG` file, which means that git ignores it automatically, according to this informal specification: https://bford.info/cachedir/
The same is true for other directories that I have such as `.mypy_cache` and `.ruff_cache`.

ayushjariyal:
@danielhollas , Thanks for the reviewing the PR!

I created the virtual environment using the command:

`python -m venv env_name`

And yes, when I run `git status`, the virtual environment folder shows up as an untracked file.
Screenshot from 2025-04-20 19-22-21

I use the same command to create the virtual environment for `aiida-core`, and there it's not an issue because `/venv*/` is already included in the `.gitignore` file.

I don't current use `uv`, but if it handles environment folders with the `CACHEDIR.TAG`, that sounds convenient. I'll give  it a try and consider using it for future environments.

danielhollas:
> containing a CACHEDIR.TAG file, which means that git ignores it automatically, 

I was wrong about this. git ignores uv's `.venv` folder because the folder itself contains a `.gitignore` file with `*` on the first line.

> I don't currently use uv, but if it handles environment folders with the CACHEDIR.TAG, that sounds convenient. I'll give it a try and consider using it for future environments.

I highly recommend giving it a try, `uv` is so much faster than `pip`. :rocket: Usual workflow is

```
uv venv
uv pip install -e .[dev]
source .venv/bin/activate
```

Then you can activate the environment as usual.

PR #474, opened 2025-04-26
Replacing `pkg_resource` with `importlib.resources`

### LINKED ISSUE #432: Replace pkg_resource with importlib
`pkg_resources` module (distributed with setuptools) is deprecated, and users are encouraged to use `importlib.resources` and `importlib.metadata`. See https://setuptools.pypa.io/en/latest/pkg_resources.html

So we currently have undeclared dependency on setuptools (currently satisfied transitively via `dulwich` and `jsonschema`)

ayushjariyal:
Closes #432


ayushjariyal:
@danielhollas, can you please review these changes. I am not able to understand that why these tests are failing.

aiidalab/registry/core.py
```diff
@@ -34,9 +34,9 @@ def from_package(cls):
         return cls(
             **{
                 field.name: json.loads(
-                    pkg_resources.resource_string(
-                        "aiidalab.registry", f"schemas/{field.name}.schema.json"
-                    )
+                    resources.files("aiidalab.registry.schemas")
+                    .joinpath(f"{field.name}.schema.json")
```
danielhollas, line 38: It looks like the following fixes the tests for python 3.9

```suggestion
                    resources.files("aiidalab.registry")
                    .joinpath(f"schemas/{field.name}.schema.json")
```

I suspect the problem was that the `schemas` folder does not contain the `__init__.py` file so it's not a package from python's point of view.

with `uv` you can easily specify python version when creating the venv like this

```
uv venv -p 3.9
```

Here's the error that you get on python 3.9

```python
❯ python
Python 3.9.21 (main, Dec 19 2024, 14:33:53) 
[Clang 18.1.8 ] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> from importlib import resources
>>> resources.files('aiidalab.registry')
PosixPath('/home/hollas/atmospec/aiidalab/aiidalab/registry')
>>> resources.files('aiidalab.registry.schemas')
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/home/hollas/.local/share/uv/python/cpython-3.9.21-linux-x86_64-gnu/lib/python3.9/importlib/resources.py", line 147, in files
    return _common.from_package(_get_package(package))
  File "/home/hollas/.local/share/uv/python/cpython-3.9.21-linux-x86_64-gnu/lib/python3.9/importlib/_common.py", line 14, in from_package
    return fallback_resources(package.__spec__)
  File "/home/hollas/.local/share/uv/python/cpython-3.9.21-linux-x86_64-gnu/lib/python3.9/importlib/_common.py", line 18, in fallback_resources
    package_directory = pathlib.Path(spec.origin).parent
  File "/home/hollas/.local/share/uv/python/cpython-3.9.21-linux-x86_64-gnu/lib/python3.9/pathlib.py", line 1082, in __new__
    self = cls._from_parts(args, init=False)
  File "/home/hollas/.local/share/uv/python/cpython-3.9.21-linux-x86_64-gnu/lib/python3.9/pathlib.py", line 707, in _from_parts
    drv, root, parts = self._parse_args(args)
  File "/home/hollas/.local/share/uv/python/cpython-3.9.21-linux-x86_64-gnu/lib/python3.9/pathlib.py", line 691, in _parse_args
    a = os.fspath(a)
TypeError: expected str, bytes or os.PathLike object, not NoneType
```

danielhollas (changes requested):
@ayushjariyal great, tests are passing now!

Let's remove the last vestiges of pkg_resources in `aiidalab/registry/web.py` as part of this PR.

ayushjariyal:
@danielhollas , Done!

aiidalab/registry/web.py
```diff
@@ -35,28 +34,23 @@ def copy_static_tree_from_path(base_path, static_path):
 
 
 def _walk_pkg_resources(package, root):
-    paths = pkg_resources.resource_listdir(package, root)
-    for path in paths:
```
danielhollas, line 39: Huh, it looks like you actually fixed a bug here where we would iterate over some directories twice before?
I'll verify a bit more, but the new implementation looks correct. I'll try to add a regression test for this.

danielhollas (approved):
Nice, thanks you!

I'll try to add a new test before merging this since the aiidalab.registry package is tested rather sparsely.

danielhollas (changes requested):
Hi @ayushjariyal,

apologies I did not get to this due to other work. If you'd like to push this forward, we'd need a bit more testing to ensure correctness of these changes, since the current testing of this code is very crude.

At the very least, we need unit tests for `copy_static_tree_from_package` and `_walk_pkg_resources`.

aiidalab/registry/web.py
danielhollas: When writing a unit test for this, I would recommend printing `paths` here and comparing the old and new versions, they are not the same! (although the new one seems more correct as it doesn't duplicate the paths).

edan-bainglass:
Hi @ayushjariyal. Do you still plan to finalize this one? Great if we can merge it soon 🙂

ayushjariyal:
Hi @edan-bainglass, thanks for checking in! I'm still interested in getting this merged.

I haven't been actively following the project recently, so I'm not very familiar with the latest changes in the codebase. If the PR looks good from your side and can be merged as is, that would be great. Otherwise, let me know what's needed and I'll take a look.

Thanks!

edan-bainglass:
> Hi @edan-bainglass, thanks for checking in! I'm still interested in getting this merged.
> 
> I haven't been actively following the project recently, so I'm not very familiar with the latest changes in the codebase. If the PR looks good from your side and can be merged as is, that would be great. Otherwise, let me know what's needed and I'll take a look.
> 
> Thanks!

Thanks @ayushjariyal. You can refer to the last few comments in this PR for the remaining requirements. @danielhollas can provide further info if needed 🙂

Issue #343, opened 2022-12-14
Simplify package setup.


yakutovicha:
- [ ] Replace `setup.cfg/setup.py` with `pyproject.toml` (see https://setuptools.pypa.io/en/latest/userguide/pyproject_config.html)
- [ ] Replace setuptools with flit (see how aiidateam/aiida-core pyproject.toml is configured)

danielhollas:
@ayushjariyal this might be a good issue to tackle if you want to learn more about python packaging.
It also pairs well with #432.

ayushjariyal:
@danielhollas,Thank you so much for the suggestion! 
I’d love to take this up — it sounds like a great opportunity to deepen my understanding of Python packaging, and I’m excited that it complements #432 as well.
I’ll start looking into it right away and share any questions or thoughts I have as I go along. Really looking forward to working on this!

danielhollas:
This was done in #475

PR #475, opened 2025-04-27
Replace `setup.cfg`/`setup.py` with `pyproject.toml`

ayushjariyal:
issue #343


ayushjariyal:
@danielhollas ,@yakutovicha , can you please review this PR?

pyproject.toml
```diff
@@ -1,7 +1,104 @@
...
+    "toml~=0.10",
+    "traitlets~=5.0",
+    "watchdog~=4.0",
+]
+description = "Implements core functions of AiiDAlab"
+license = {file = 'LICENSE'}
+readme = 'README.md'
+requires-python = '>=3.9'
+
+[project.script]
```
danielhollas, line 38: I tested by running `uv build` and got

```console
$ uv build
Building source distribution...
Unexpected names under [project]: script
Unexpected names under [project]: script
Building wheel from source distribution...
Unexpected names under [project]: script
Unexpected names under [project]: script
Successfully built dist/aiidalab-24.9.0.tar.gz
Successfully built dist/aiidalab-24.9.0-py3-none-any.whl
```

There's a typo here:

```suggestion
[project.scripts]
```

I am surprised that the tests are passing, since the `aiidalab` cli command is not available due to this mistake. We'll need to expand the test to catch this.

danielhollas (changes requested):
@ayushjariyal thanks! Looks great in general. Please also remove the `setup.py` and `setup.cfg` files.

Can I ask how did you test this?

Can I ask you to verify that the built distributions contains the exact same files as before? Specifically, you can use `uv build`, which will build a wheel and sdist and put them in a `dist/` directory. Can you check the contents of the resulting archive `dist/dist/aiidalab-24.9.0.tar.gz` and verify it against the distribution built from the main branch?

danielhollas:
I've pushed a commit that adds a basic [smoke test](https://en.wikipedia.org/wiki/Smoke_testing_(software)) that verifies that the `aiidalab` command works, which uncovers the typo in `pyproject.toml`

ayushjariyal:
Hi @danielhollas, Thanks for the feedback.

At the time of submission, I mainly made the changes following how it's done in the `aiida-core` module, and i hadn't fully tested everything yet.
After your comment, I tested it properly:

> I used `uv build` to build the distributions.

> I checked the contents of `dist/aiidalab-24.9.0.tar.gz` and compare it with the one from the main branch. the files seem to match.

> I also ran the test now, and they all are passing except `test_cli.py` file.

I have also removed the `setup.py` and setup.cfg` files as you asked.
Please let me know if i missed anything or if something else needs to be done!

setup.cfg
```diff
@@ -1,94 +0,0 @@
...
-    pydata-sphinx-theme~=0.13.3
-    sphinx~=4.5.0
-    sphinx-design~=0.4.0
-    sphinxcontrib-contentui
-    sphinxcontrib-details-directive;python_version>='3.0'
-    # NOTE: These need to be pinned due to incompatibility with sphinx<5.0
-    # https://github.com/sphinx-doc/sphinx/issues/11890
-    # We can't easily upgrade to sphinx v5.0 since sphinx-panels is incompatible
-    # and is deprecated, we'd need to migrate to sphinx-design.
-    # https://sphinx-design.readthedocs.io/en/latest/get_started.html#migrating-from-sphinx-panels
```
danielhollas, line 68: Can you please move this comment to pyproject.toml?

setup.cfg
```diff
@@ -1,94 +0,0 @@
...
-    sphinxcontrib-serializinghtml==1.1.5
-
-[mypy]
-strict = True
-ignore_missing_imports = True
-strict_optional = True
-show_error_codes = True
-
-[mypy-aiidalab.registry.*]
-ignore_errors = True
```
danielhollas, line 82: This needs to be moved to pyproject.toml

danielhollas (changes requested):
@ayushjariyal thanks! Almost there, thanks! 

The pre-commit mypy hook is now failing, can you please fix its configuration in `.pre-commit-config.yaml`, the `--config-file` parameter needs to point to `pyproject.toml` instead of `setup.cfg`. Please verify by running `pre-commit run --all-files` locally.

Also please remove file `MANIFEST.in`, it is specific to setuptools and is not needed anymore.

setup.cfg
ayushjariyal: After adding this to the `pyproject.toml `file and running the tests, it shows the following failure message:
`pyproject.toml: [mypy]: Unrecognized option: aiidalab.registry.* = {'ignore_errors': True}`

danielhollas:
Please check the mypy docs, the syntax in pyproject.tomp is probably
slightly different.

setup.cfg
ayushjariyal: Done!

setup.cfg
ayushjariyal: Done!

ayushjariyal:
@danielhollas , Done!

danielhollas (approved):
LGTM, thanks for working on this @ayushjariyal!

PR #489, opened 2026-02-12
Pin setuptools

danielhollas:
setuptools 82.0 removed the deprecated `pkg_resources` modules. https://setuptools.pypa.io/en/latest/history.html#v82-0-0 Until we migrate, we need to pin to an older versions.

The `pkg_resources` migration is tracked in #432, and there's an open PR #474


yakutovicha (approved):
thanks, @danielhollas!

PR #494, opened 2026-03-24
Add ipywidgets to core-packages

danielhollas:
In aiidalab we have a concept of core packages which must never be re-installed in the image. Up till now that was just `aiida-core` and `jupyter-client`. Before we even show an app on Appstore page, which check that a given app version does not have package requirements that would conflict with the core packages, and if they do, we don't show the version at all to the user. Let's add ipywidgets to this list, so that users don't try to install older package versions to the new nbclassic-based image. (note that we already pin ipywidgets in the image, so even if the user tried the installation, it would ultimately fail). 

## Test plan

The following has been tested in the nbclassic image (`ghcr.io/aiidalab/full-stack:pr-479`)

### Before this PR

Before this PR, in the Appstore one can try to install AWB 2.5.0

image

The installation itself fails since we pin the ipywidgets version to v8 in the nbclassic docker image.

image

### On this PR

image

### DIFF
--- a/aiidalab/app.py
+++ b/aiidalab/app.py
@@ -59,7 +59,11 @@ if TYPE_CHECKING:
 
 logger = logging.getLogger(__name__)
 
-_CORE_PACKAGES = [Package("aiida-core"), Package("jupyter-client")]
+_CORE_PACKAGES = [
+    Package("aiida-core"),
+    Package("jupyter-client"),
+    Package("ipywidgets"),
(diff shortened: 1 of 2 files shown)


edan-bainglass (approved):
Good add. LGTM!

danielhollas:
Tested that the new AWB pre-release now appears in appstore on this branch

image

PR #488, opened 2026-02-12
Add citation handler

edan-bainglass:
This PR adds a citation handler to the app base page:

```
    {%if metadata and metadata.citation %}
    <h2>
        Citation
    </h2>
    <div id='citation'>
        {{ metadata.citation }}
    </div>
    {% endif %}
```

as well as to metadata.py (`Metadata` dataclass and parser). This allows the following:

image

Issue #492, opened 2026-02-26
Flaky tests


danielhollas:
They are some flaky tests, we should fix them

One of them is

```
=================================== FAILURES ===================================
_______________________________ test_prereleases _______________________________

generate_app = <function generate_app.<locals>._generate_app at 0x7f8be08a6ac0>

    def test_prereleases(generate_app):
        app = generate_app()
    
        # without prereleases tick
        app.refresh()
>       assert app.has_prereleases
E       assert False
E        +  where False = <aiidalab.app.AiidaLabApp object at 0x7f8be0a69550>.has_prereleases
```

danielhollas:
Did some investigation here. One can induce the test flakes much faster if the allowed frequency of calls to `refresh` is increased (in `aiidalab/app.py`), e.g.

```diff
diff --git a/aiidalab/app.py b/aiidalab/app.py
index 059e26f..18947cb 100644
--- a/aiidalab/app.py
+++ b/aiidalab/app.py
@@ -970,7 +970,7 @@ class AiidaLabApp(traitlets.HasTraits):
         else:
             self.dependencies_to_install = []
 
-    @throttled(calls_per_second=1)
+    @throttled(calls_per_second=100)
     def refresh(self) -> None:
         """Refresh app state."""
         with self._show_busy():
```

And then running the test in a loop, it usually crashed within first 10 tries

```bash
i=0; while pytest -s tests/test_appclass.py::test_prereleases; do let i++; echo "iteration $i"; done
```

I think the fundamental problem here is that the `refresh` function is not thread safe and we don't have a lock around it. The `throttled` decorator adds a superficial lock around it, but if the two calls happen to be within the throttled frequency (and the previous refresh is not finished yet) I think we have a problem.

Also, I tested that if I remove the call to `refresh_async` in the __init__ the flakiness disappears. So there's definitely something wrong with the threading.

```diff
❯ git diff aiidalab/app.py 
diff --git a/aiidalab/app.py b/aiidalab/app.py
index 059e26f..07ba1df 100644
--- a/aiidalab/app.py
+++ b/aiidalab/app.py
@@ -803,7 +803,7 @@ class AiidaLabApp(traitlets.HasTraits):
 
         self.is_installed = self._app.is_installed
         self.path = str(self._app.path)
-        self.refresh_async()
+        #self.refresh_async()
 
         self._watch = None
         if watch:
```

PR #426, opened 2024-04-30
App install: Handle invalid Git URLs better

danielhollas:
Closes #396

PR #499, opened 2026-04-10
Add `citations` key to `metadata` schema

edan-bainglass:
Missed in #488

### Changes

- [x] Add `citations` entry to metadata schema
- [x] Fix typo in `citations` extraction from metadata
- [x] Fix schema URLs to upcoming 26.04.0 release

@danielhollas tests won't pass, as there is no 26.04.0 release, we need to merge and release 26.04.0. IMO, this system is far from ideal. Unclear what the rational was behind it. Let's discuss to see if it can be improved.

Note that I am still trying to figure out how to test this locally. Need to go through the docs.

### DIFF
--- a/aiidalab/metadata.py
+++ b/aiidalab/metadata.py
@@ -94,7 +94,7 @@ def _parse_setup_cfg(
         categories = [c for c in categories.split("\n") if c]
     yield "categories", categories
 
-    yield "citations", metadata_pep426.get("citation", "").splitlines()
+    yield "citations", metadata_pep426.get("citations", "").splitlines()
 
 
 @dataclass
--- a/aiidalab/metadata.py
+++ b/aiidalab/metadata.py
@@ -94,7 +94,8 @@ def _parse_setup_cfg(
         categories = [c for c in categories.split("\n") if c]
     yield "categories", categories
 
-    yield "citations", metadata_pep426.get("citations", "").splitlines()
+    citations = aiidalab.get("citations", metadata_pep426.get("citations", ""))
+    yield "citations", citations.splitlines()
 
 
 @dataclass
--- a/aiidalab/registry/schemas/app.schema.json
+++ b/aiidalab/registry/schemas/app.schema.json
@@ -1,6 +1,6 @@
 {
     "$schema": "http://json-schema.org/draft-07/schema#",
-    "$id": "https://raw.githubusercontent.com/aiidalab/aiidalab/v26.04.0/aiidalab/registry/schemas/app.schema.json",
+    "$id": "https://raw.githubusercontent.com/edan-bainglass/aiidalab/refs/heads/fix-citations-handler/aiidalab/registry/schemas/app.schema.json",
     "$ref": "#/definitions/App",
     "definitions": {
         "App": {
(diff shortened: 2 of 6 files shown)


danielhollas:
> @danielhollas tests won't pass, as there is no 26.04.0 release, we need to merge and release 26.04.0. IMO, this system is far from ideal. Unclear what the rational was behind it. Let's discuss to see if it can be improved.

I would like to see the tests pass here at least once before merging. For that you could point the schemas to main branch (or this branch, if there are fixes here) and then we switch just before merging? Agreed that this seems far from ideal :-)

edan-bainglass:
> I would like to see the tests pass here at least once before merging. For that you could point the schemas to main branch (or this branch, if there are fixes here) and then we switch just before merging? Agreed that this seems far from ideal :-)

Done! All tests pass!

Interestingly `tests/test_appclass.py::test_prereleases` failed on 3.12 in the first attempt - passed on a retry. Possibly flaky. Something to keep in mind (~will open an issue~ you already did [here](https://github.com/aiidalab/aiidalab/issues/492#issue-3997190233) 👍).

If no further comments, let's get this one merged (after I revert the schema references), then do a release 🙏

danielhollas:
#492

edan-bainglass:
> #492

Right 👍 Linked in my comment 🙂

edan-bainglass:
@danielhollas ping me when you're good with this. I'll then revert the schema pointers and we merge/release 🙏

danielhollas (approved):
Thanks! I'm good with this. 

Before making a release, I'd like to get in a small bugfix #426. I still need to test it but otherwise should be ready to review.

edan-bainglass:
@danielhollas how about permanently pinning the schemas to `main`? Wouldn't this be easiest?

danielhollas:
> @danielhollas how about permanently pinning the schemas to main? Wouldn't this be easiest?

Not a fan, because merging a problematic commit here could bring down the registry, because the version pinning that we do in `aiidalab-registry` would no longer work. Let's keep the version pinning as is now (i.e. pin to future version, and accept that the test will fail, and make a release). We can resolve this in the future (or switch to something completely different like pydantic schemas?). But I don't feel strongly here, do what you think is best.

edan-bainglass:
> Not a fan, because merging a problematic commit here could bring down the registry, because the version pinning that we do in `aiidalab-registry` would no longer work. Let's keep the version pinning as is now (i.e. pin to future version, and accept that the test will fail, and make a release). We can resolve this in the future (or switch to something completely different like pydantic schemas?). But I don't feel strongly here, do what you think is best.

I'll restore the future version pin. We can discuss this setup at a later time 🙂

danielhollas:
> @danielhollas how about permanently pinning the schemas to main?

So I have a counterproposal but in a similar vein: instead of pinning to the tag, what if we pin to a specific commit? That doesn't fully resolve the chicken-and-egg problem, but I think it does make it slightly better, since then you can have a green CI within the same PR. Pinning to a commit is also inherently safer since tags are mutable and can be deleted. Obviously still not great, but perhaps a bit of an improvement? Ideally this also needs to be documented, somewhere in the code maybe? (I guess problem with json is it can't have comments?)

PR #501, opened 2026-04-10
Only install compatible packages during app update

### LINKED ISSUE #500: Update button doesn't filter out incompatible versions
When testing the old image with AWB 2.4.0, I see this

Image

However, if I click update, the app tries to update all to way to just released 3.0.0a0, which is incompatible (due to ipywidgets) 

~~and is a pre-release (and notice I didn't have pre-release checkbox ticked!~~ I was wrong, only happens if I tick pre-release box.

Working on a fix

danielhollas:
This was apparently an oversight when we implemented the filtering of packages based on whether they satisfy core dependencies (e.g. aiida-core).

When we run update, we passed `version=None` to `_AiidaLabApp.install()` in which case that method was supposed to pick the latest available version, but it was not using the proper filtering.

Closes #500 

## Test plan

See issue #500. Install AWB 2.4.0 first, and then go to App management page (http://localhost:8888/apps/apps/home/single_app.ipynb), tick the "pre-releases" box and click "Update button". The Update should pick up version 2.5.0 on this PR, while it picks 3.0.0a0 on main.

### DIFF
--- a/aiidalab/app.py
+++ b/aiidalab/app.py
@@ -537,10 +537,10 @@ class _AiidaLabApp:
         post_install_triggers: bool = True,
     ) -> None:
         if version is None:
-            versions = sort_semantic(self.releases, prereleases=prereleases)
-            if len(versions) == 0:
+            try:
+                version = next(self.available_versions(prereleases=prereleases))
+            except StopIteration:
                 raise ValueError(f"No versions available for '{self}'.")
-            version = versions[0]
         if python_bin is None:
             python_bin = sys.executable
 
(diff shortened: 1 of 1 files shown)


aiidalab/app.py
```diff
@@ -900,8 +900,6 @@ def update_app(
             # Installing with version=None automatically selects latest
             # available version.
             version = self.install_app(version=None, stdout=stdout)
-            FIND_INSTALLED_PACKAGES_CACHE.clear()
-            self.refresh()
```
danielhollas, line 904: This is already called in `self.install_app` so we were refreshing twice needlessly.

aiidalab/app.py
```diff
@@ -537,10 +537,10 @@ def install(
         post_install_triggers: bool = True,
     ) -> None:
         if version is None:
-            versions = sort_semantic(self.releases, prereleases=prereleases)
```
danielhollas, line 540: `self.releases` holds all versions that we found, whereas `self.available_versions` filters them based on compatibility.

edan-bainglass (approved):
LGTM!

PR #507, opened 2026-05-13
Fix citations metadata handling

edan-bainglass:
@danielhollas had to add a guard around the citation handler due to a failure. Now, the failure is actually due to the QE app still having the old style format (string style) and was crashing the build. Even though I have since PR'd the updated format there, I think this guard is good in general to have.


edan-bainglass:
@danielhollas force merged for testing. We can revert if you have any critical issues.

PR #508, opened 2026-05-19
CI: Test with lowest version of dependencies

danielhollas:
When aiidalab package is being installed into the docker stack image, many of its dependencies are already present in the image in the conda environment, and are therefore not installed at their highest (most up-to-date) versions. It's therefore pertinent to make sure that our lowest version pins are accurate, which we can achieve by using the uv's 'lowest-direct' resolution algorithm. We have the exact same setup in AWB repo.

Bumps minimum versions of several packages to make the new jobs pass.


edan-bainglass (approved):
LGTM!

PR #510, opened 2026-05-28
Add citations property to `AiidaLabApp` class

edan-bainglass:
Extracted from app metadata, which routes through the setup.cfg aiidalab section.
This is used, for example, in the home app's app manager to display app citations.


danielhollas (approved):
Neverending story :-D

PR #511, opened 2026-05-28
Add typing to aiidalab.registry module

danielhollas:
This adds a minimum-effort typing to the `aiidalab.registry` module, hopefully making future changes a bit safer.


aiidalab/registry/yaml.py
```diff
@@ -12,7 +15,7 @@
 class JsonYamlLoader(jsonref.JsonLoader):  # type: ignore[misc]
     safe_yaml = YAML(typ="safe")
 
-    def __call__(self, uri, **kwargs):
+    def __call__(self, uri: str, **kwargs: Any) -> Any:
```
edan-bainglass, line 18: Is it worth typing kwargs, or rather, is it standard practice?

aiidalab/utils.py
```diff
@@ -78,9 +78,12 @@ class PEP508CompliantUrl(str):
     pass
 
 
+_ParseAppCallable = Callable[[str], dict[str, Any]]
+
+
 def parse_app_repo(
     url: str, metadata_fallback: dict[str, Any] | None = None
-) -> dict[str, Any] | None:
+) -> dict[str, Any]:
```
edan-bainglass, line 88: So not really optional?

aiidalab/utils.py
```diff
@@ -122,7 +125,7 @@ class throttled:  # noqa: N801
 
     def __init__(self, calls_per_second: int = 1):
         self.calls_per_second = calls_per_second
-        self.last_start = defaultdict(lambda: -1)  # type: ignore
+        self.last_start = defaultdict(lambda: -1.0)  # type: ignore
```
edan-bainglass: Is the type ignore still required then? I guess it's for something else?

edan-bainglass (approved):
LGTM! Left a few comments/questions.

aiidalab/registry/yaml.py
danielhollas: mypy would otherwise complained about untyped function definition (and more importantly, unless the function definition is fully typed, mypy doesn't check it at all by default)

aiidalab/utils.py
danielhollas: Mypy wanted an explicit type annotation, so I added one and removed the type ignore

aiidalab/utils.py
danielhollas: Looking at the function code it's clear that it never returns None so this annotation was simply wrong (and was causing problems elsewhere)

aiidalab/registry/yaml.py
edan-bainglass: Interesting

aiidalab/utils.py
edan-bainglass: I'm confused. Looks like the type ignore is still there? And no type annotation?

danielhollas: You were probably looking at the outdated version of the code

image

PR #512, opened 2026-05-29
CI: Add registry build job

danielhollas:
We should know immediately in case the registry build broke.

(in the past I think we had a situation where we made a release and only then realizes it broke the build)


edan-bainglass (approved):
LGTM!

PR #513, opened 2026-05-29
Fix test flakes

### LINKED ISSUE #492: Flaky tests

danielhollas:
Fixes #492, see https://github.com/aiidalab/aiidalab/issues/492#issuecomment-4576132529 for the debugging exploration.

I think the fundamental issue here is the the `app.refresh` method is not threadsafe and we don't have a lock around it. 

This PR is focused only on fixing the test flakiness, by making sure that the background thread that runs from the `refresh_async` method finishes before the `generate_app` fixture is ready. 

I think the underlying issue of the thread safety still should be fixed, I'll open a separate issue about that since I am a bit scared to touch this code and it will need careful testing.

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -1,4 +1,5 @@
 import sys
+import threading
 from pathlib import Path
 
 import pytest
@@ -48,6 +49,11 @@ def generate_app(monkeypatch, app_registry_path):
         monkeypatch.setattr(_AiidaLabApp, "is_installed", lambda _: True)
         monkeypatch.setattr(_AiidaLabApp, "is_registered", lambda _: True)
         app = AiidaLabApp(name, app_data, aiidalab_apps_path, watch=watch)
+
+        # The app refresh is running in a background thread
+        # let's wait till it finishes
+        while app.busy or threading.active_count() > 1:
+            pass
 
         return app
 
--- a/tests/test_appclass.py
+++ b/tests/test_appclass.py
@@ -11,9 +11,8 @@ from aiidalab.app import AiidaLabApp, AiidaLabAppWatch
 
 def test_init_refresh(generate_app):
     app = generate_app()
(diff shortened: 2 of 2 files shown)


tests/conftest.py
```diff
@@ -49,6 +50,11 @@ def _generate_app(
         monkeypatch.setattr(_AiidaLabApp, "is_registered", lambda _: True)
         app = AiidaLabApp(name, app_data, aiidalab_apps_path, watch=watch)
 
+        # The app refresh is running in a background thread
```
danielhollas, line 53: This thread is started from the `__init__` method here:

https://github.com/aiidalab/aiidalab/blob/abcceea4582faaab635a64f5fe1e99e47def63a3/aiidalab/app.py#L806

tests/test_appclass.py
```diff
@@ -11,9 +11,8 @@
 
 def test_init_refresh(generate_app):
     app = generate_app()
-    assert len(app.available_versions) == 0
```
danielhollas, line 14: This condition was inherently racy! If the background thread was fast enough, the `available_versions` would already be refreshed.

edan-bainglass (approved):
LGTM! FYI, we will discuss next week how much of AiiDAlab survives the future plans, so "careful" how much time you're spending here 😅

PR #515, opened 2026-06-02
Uninstall python package during app removal

### LINKED ISSUE #405: `aiidalab uninstall` does not remove the AiiDALab app package installed by pip
In the test, the `aiidalab-widgets-base` is installed by `aiidalab install` and uninstall with `aiidalab uninstall`. It will cause the problem that the app is deleted but the package still in the python path. Which potentially lead to the fail of `pip check`, when the aiidalab-widgets-base has newer version i.e `2.2.0` but not yet bump in `aiidalab-qe` i.e `2.1.0`. 
It will result the: "aiidalab-qe 23.10.x has requirement aiidalab-widgets-base==2.1.0, but you have aiidalab-widgets-base 2.2.0."

It caused from `aiidalab uninstall` only move the app to trash but not run `pip uninstall`, so the fixture `generate_aiidalab_install_output` is not roll back the environment. 

The ultimate solution is install the awb from `pip install` since it will no more be the app (https://github.com/aiidalab/aiidalab-widgets-base/issues/550).

danielhollas:
Currently, when an app is being uninstalled, the only thing we do is to remove the app directory (the git repo in `/home/jovyan/apps/<app-name>`). We don't remove the apps python package (if any) or any of its python dependencies. 

Removing apps python dependencies would be unsafe, since they can be shared with other apps, but we can safely remove the python package of the app itself, since keeping it around is (at the very least) confusing. This is what this PR implements. 

Note, this is only done on a best-effort basis, since the app name might not be the same as python package name. Notably, this is the case for QeApp, where the app name is `quantum-espresso`, while the python package is `aiidalab-qe`. Perhaps we can special case QeApp here?

Closes #405 

### Test plan

1. Tested in CI on aiidalab-docker-stack repo: https://github.com/aiidalab/aiidalab-docker-stack/pull/571

2. Manual testing

```console
$ aiidalab install --yes aurora

$ pip show aurora
Name: aurora
Version: 0.12.3
Summary: An AiiDAlab application for the Aurora BIG-MAP Stakeholder initiative.
Home-page: https://github.com/EmpaEconversion/aiidalab-aurora
Author: Edan Bainglass, Francisco F. Ramirez, Loris Ercole, Giovanni Pizzi
Author-email: edan.bainglass@psi.ch
License: 
Location: /home/jovyan/.local/lib/python3.9/site-packages
Requires: aiida-aurora, aiidalab, dgbowl-schemas, ipyfilechooser, ipympl, ipywidgets, numpy, pandas, pydantic, seaborn
Required-by:
$ aiidalab uninstall --yes aurora
Collecting apps to uninstall... Done.

Would uninstall:

  App name    Version    Path
  ----------  ---------  ------------------------
  aurora      0.12.3     /home/jovyan/apps/aurora

  ▬:detached •:modified

Uninstalled 'aurora' ('/home/jovyan/apps/aurora').
$ pip show aurora
WARNING: Package(s) not found: aurora
```

### DIFF
--- a/aiidalab/app.py
+++ b/aiidalab/app.py
@@ -324,10 +324,13 @@ class _AiidaLabApp:
             or self.path.joinpath("pyproject.toml").is_file()
         ):
             logger.info(f"Running 'pip uninstall --user {self.name}'")
-            process = run_pip_uninstall(str(self.path), python_bin=python_bin)
-            for line in io.TextIOWrapper(process.stdout, encoding="utf-8"):
-                logger.warning(line)
+            process = run_pip_uninstall(str(self.name), python_bin=python_bin)
             process.wait()
+            for line in io.TextIOWrapper(process.stdout, encoding="utf-8"):
+                if process.returncode == 0:
+                    logger.info(line)
+                else:
+                    logger.warning(line)
             if process.returncode != 0:
                 msg = f"pip failed to uninstall python package {self.name}"
                 logger.warning(msg)
--- a/aiidalab/app.py
+++ b/aiidalab/app.py
@@ -323,11 +323,13 @@ class _AiidaLabApp:
             self.path.joinpath("setup.py").is_file()
             or self.path.joinpath("pyproject.toml").is_file()
         ):
-            logger.info(f"Running 'pip uninstall --user {self.path}'")
+            logger.info(f"Running 'pip uninstall --user {self.name}'")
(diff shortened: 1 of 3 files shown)


danielhollas:
This proved a bit trickier then I thought. I think the current code works, but requires more testing so keeping in draft and moving to other more pressing things for now. No need to review yet.

PR #516, opened 2026-06-03
Move setuptools dep into registry extras

danielhollas:
`pkg_resources`, which is a subpackage of setuptools, is only used in the `aiidalab.registry` submodule so we can safely move it to the `registry` extras. 

This change makes the in-progress PR that removes this dependency (#474) much less pressing.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -24,7 +24,6 @@ dependencies = [
     "pip>=25.2",
     "requests~=2.27",
     "requests-cache~=1.0",
-    "setuptools>=68.0,<82.0",
     "tabulate~=0.9",
     "toml~=0.10",
     "traitlets~=5.0",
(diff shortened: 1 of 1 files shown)


pyproject.toml
```diff
@@ -55,6 +54,7 @@ dev = [
     "types-click-spinner~=0.1",
 ]
 registry = [
+    "setuptools>=65.0,<82.0",
```
danielhollas: In #508 I've set the minimum setuptools version somewhat randomly to 68, but that turned out to be problematic, because in the current python 3.9 image we have setuptools 65.4 install via conda, and when I was testing #515 and installed aiidalab in editable mode via pip, it triggered a setuptools reinstall, which resulted in much sadness.  

EDIT: Nevermind, downgrading the minimum versions makes the registry test fail in CI here. :-(

edan-bainglass (approved):
LGTM!

yakutovicha (approved):
second approval :)

PR #521, opened 2026-07-30
Add py.typed marker, fix some type annotations

danielhollas:
While I was looking into adding type annotations to aiidalab-home, I noticed 2 things:

1. aiidalab package was not declaring itself to be type-annotated (per [PEP 561](https://peps.python.org/pep-0561/#packaging-type-information)
2. Two type annotations were wrong (not allowing None values)


edan-bainglass (approved):
Good catch. LGTM!

PR #522, opened 2026-07-30
Update to ruff 0.16.1, use only default ruff rules

danielhollas:
ruff 0.16.0 hugely expanded its default rule set from 59 to 413 rules. I propose to use this new default and simplify our current ruff configuration, since there is anyway a big overlap with what we had enabled, and I generally trust the astral team to do a good job of picking up important rules. 

See https://astral.sh/blog/ruff-v0.16.0#better-default-rule-set for more details. 

@yakutovicha @edan-bainglass if this sounds like a good plan, I'll do the same for the other repos.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -13,11 +13,11 @@ repos:
           - id: trailing-whitespace
 
     - repo: https://github.com/astral-sh/ruff-pre-commit
-      rev: v0.15.20
+      rev: v0.16.1
       hooks:
           - id: ruff-format
             exclude: ^docs/.*
-          - id: ruff
+          - id: ruff-check
             args: [--fix, --exit-non-zero-on-fix, --show-fixes]
 
     - repo: https://github.com/jumanjihouse/pre-commit-hook-yamlfmt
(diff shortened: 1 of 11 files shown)


edan-bainglass:
> @yakutovicha @edan-bainglass if this sounds like a good plan, I'll do the same for the other repos.

I support this but also am concerned it may be a headache? Unless we're great and have been doing things the right way 🤓 Otherwise, would Ruff fix automatically, or will this be "major" work? @yakutovicha if I understand correctly, it is precisely these sort of "bugs" that make back-porting anything a nightmare 😅

yakutovicha (approved):
I like it. Also, the code seems to have become simpler.

danielhollas:
> I support this but also am concerned it may be a headache? Unless we're great and have been doing things the right way

That's a fair point, but just to be clear, the current settings in this repo enable 322 rules so it's not such a big jump. There are not that many changes in this PR and most of them were autofixed by ruff. And if there's a particular rule that would be too noisy for a given repo we can always disable it.

edan-bainglass (approved):
LGTM! Thanks @danielhollas. Regarding my comment, I thought we were at 59 jumping to 413 😅 Good to know we were already checking most.

PR #525, opened 2026-08-15
Discard str-casting of metadata values

### LINKED ISSUE #524: Issue with citation rendering
There appears to be an issue with the citation reader in the app store 🤔 I can't tell yet if it is related to the new stack, because this feature is in a version of the home app newer than that deployed on the demo server.

See comment by @danielhollas [here](https://github.com/aiidalab/aiidalab/pull/510#pullrequestreview-4378130926)

---

image

---

Until I added citations (#488 and related), it was assumed that all entries are string based: (from aiidalab/app.py)
```python
    def _get_from_metadata(self, what: str) -> str:
        """Get information from metadata."""
        try:
            return f"{self._app.metadata[what]}" <-------------------- 🥲
        except KeyError:
            return f'Field "{what}" is not present in app metadata.'
```
That's turning a citations list of dictionaries into a string. So instead of iterating over the list, it is iterating over the chars, yielding a lot of garbage 🤦🏻‍♂️

Works without the f-string wrapper

---

image

---

Question is - why was the wrapper introduced? So having a quick look at the metadata dataclass
```python
    title: str
    description: str
    authors: None | str = None
    state: None | str = None
    documentation_url: None | str = None
    external_url: None | str = None
    logo: None | str = None
    categories: list[str] = field(default_factory=list)
    version: None | str = None
    citations: list[SimpleCitation | StandardCitation] = field(default_factory=list)
```
suggests that it is not needed? And looks like the citations are not the only list - we also had categories. Anyhow, unclear why we were force-casting to a string.

edan-bainglass:
Some metadata values are lists and are expected as such in the template. See #524 for details on the bug.

Closes #524

### DIFF
--- a/aiidalab/app.py
+++ b/aiidalab/app.py
@@ -20,7 +20,7 @@ from pathlib import Path
 from subprocess import CalledProcessError
 from threading import Thread
 from time import sleep
-from typing import TYPE_CHECKING, Any
+from typing import TYPE_CHECKING, Any, cast
 from urllib.parse import urldefrag, urlsplit, urlunsplit
 from uuid import uuid4
 
@@ -38,7 +38,7 @@ from watchdog.observers.polling import PollingObserver
 from .environment import Environment
 from .git_util import GitManagedAppRepo as Repo
 from .git_util import git_clone
-from .metadata import Metadata
+from .metadata import Metadata, MetadataType
 from .utils import (
     FIND_INSTALLED_PACKAGES_CACHE,
     Package,
@@ -776,7 +776,7 @@ class AiidaLabApp(traitlets.HasTraits):
     def __init__(
         self,
         name: str,
-        app_data: dict[str, Any] | None,
+        app_data: dict[str, MetadataType] | None,
         aiidalab_apps_path: str | None,
         watch: bool = True,
     ):
@@ -998,33 +998,33 @@ class AiidaLabApp(traitlets.HasTraits):
         """Return metadata dictionary. Give the priority to the local copy (better for the developers)."""
         return self._app.metadata
 
-    def _get_from_metadata(self, what: str) -> Any:
+    def _get_from_metadata(self, what: str) -> MetadataType:
         """Get information from metadata."""
         try:
-            return self._app.metadata[what]
+            return cast(MetadataType, self._app.metadata[what])
         except KeyError:
             return f'Field "{what}" is not present in app metadata.'
 
     @property
     def authors(self) -> str:
-        return self._get_from_metadata("authors")
+        return cast(str, self._get_from_metadata("authors"))
 
     @property
     def description(self) -> str:
-        return self._get_from_metadata("description")
+        return cast(str, self._get_from_metadata("description"))
 
     @property
     def title(self) -> str:
-        return self._get_from_metadata("title")
+        return cast(str, self._get_from_metadata("title"))
 
     @property
     def url(self) -> str:
         """Provide explicit link to Git repository."""
-        return self._get_from_metadata("external_url")
+        return cast(str, self._get_from_metadata("external_url"))
 
     @property
     def citations(self) -> list[dict[str, Any]]:
-        return self._get_from_metadata("citations")
+        return cast(list[dict[str, Any]], self._get_from_metadata("citations"))
 
     @property
     def more(self) -> str:
--- a/aiidalab/app.py
+++ b/aiidalab/app.py
@@ -998,10 +998,10 @@ class AiidaLabApp(traitlets.HasTraits):
         """Return metadata dictionary. Give the priority to the local copy (better for the developers)."""
         return self._app.metadata
 
-    def _get_from_metadata(self, what: str) -> str:
+    def _get_from_metadata(self, what: str) -> Any:
         """Get information from metadata."""
         try:
-            return f"{self._app.metadata[what]}"
+            return self._app.metadata[what]
         except KeyError:
             return f'Field "{what}" is not present in app metadata.'
 
@@ -1023,7 +1023,7 @@ class AiidaLabApp(traitlets.HasTraits):
         return self._get_from_metadata("external_url")
 
     @property
-    def citations(self) -> str:
+    def citations(self) -> list[dict[str, Any]]:
         return self._get_from_metadata("citations")
 
     @property
--- a/aiidalab/app.py
+++ b/aiidalab/app.py
@@ -20,7 +20,7 @@ from pathlib import Path
 from subprocess import CalledProcessError
 from threading import Thread
 from time import sleep
-from typing import TYPE_CHECKING, Any, cast
+from typing import TYPE_CHECKING, Any
 from urllib.parse import urldefrag, urlsplit, urlunsplit
 from uuid import uuid4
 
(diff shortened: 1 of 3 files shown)


edan-bainglass:
@danielhollas you made type checking strict in #450, which forbids claiming to return `str` when returning a call to `_get_from_metadata(...)` requesting a field that is 100% a string. I can go one of two ways:
1. Make `_get_from_metadata` return `str | list` and explicitly `cast` to the expected value in each metadata property (better?)
2. `strict = False` 😅

**Update** - I have 1 implemented now.

danielhollas:
A much better solution would be to define a metadata TypedDict

edan-bainglass:
> A much better solution would be to define a metadata TypedDict

In addition to the `Metadata` dataclass? Seems redundant? But maybe not... 🤔

Anyhow, I've satisfied mypy, which was in no way a fun experience 🧘🏻‍♂️

edan-bainglass:
> > A much better solution would be to define a metadata TypedDict
> 
> In addition to the `Metadata` dataclass? Seems redundant? But maybe not... 🤔

Right, that's likely a better solution, but also there are other unrelated `dict[str, Any]` types that could be improved. I propose we deal with it all at once in another PR.

aiidalab/app.py
```diff
@@ -998,33 +998,33 @@ def metadata(self) -> dict[str, Any]:
...
     @property
     def url(self) -> str:
         """Provide explicit link to Git repository."""
-        return self._get_from_metadata("external_url")
+        return cast(str, self._get_from_metadata("external_url"))
 
     @property
-    def citations(self) -> str:
-        return self._get_from_metadata("citations")
+    def citations(self) -> list[dict[str, Any]]:
```
danielhollas, line 1018: I don't think this is currently accurate, because if the citations field is missing, the `_get_from_metadata` method will return a string.

danielhollas (changes requested):
I think there is still a latent bug, see my comment

I'd recommend getting rid of `_get_from_metadata` which is kind of a silly method, and use a `self._app.metadata.get(<key>, <default>)` pattern.

edan-bainglass:
@danielhollas better?

aiidalab/app.py
```diff
@@ -994,37 +994,29 @@ def refresh_async(self) -> None:
...
-        """Get information from metadata."""
-        try:
-            return f"{self._app.metadata[what]}"
-        except KeyError:
-            return f'Field "{what}" is not present in app metadata.'
-
     @property
     def authors(self) -> str:
-        return self._get_from_metadata("authors")
+        return self._app.metadata.get("authors") or "No authors provided"
```
danielhollas, line 1003: Slightly better idiom

```suggestion
        return self._app.metadata.get("authors",  "No authors provided")
```

Same also below

aiidalab/metadata.py
```diff
@@ -168,3 +168,18 @@ def parse(cls, root: Path | GitPath) -> Metadata:
                 return cls(**dict(cls._parse(path)))
 
         raise ValueError(f"Directory '{root}' does not exist.")
+
+
+class MetadataDict(TypedDict):
```
danielhollas, line 173: Not all of these fields are always present, see e.g. this code that only defines title and description: 

https://github.com/aiidalab/aiidalab/blob/78173deb0899c4eb39c043015789e7b2060d3fc5/aiidalab/app.py#L133

So we should specify `total=False` to indicate some items might be missing. We can also set `closed=True` because we don't expect any other items to be in the dictionary.
`totality` is supported only since Python 3.11, and `closed` keyword since Python 3.15 so we need to import from `typing_extensions`.

More on this: https://typing.python.org/en/latest/spec/typeddict.html

Something like this patch should work I think

```diff
diff --git a/aiidalab/metadata.py b/aiidalab/metadata.py
index 3d36394..59c171c 100644
--- a/aiidalab/metadata.py
+++ b/aiidalab/metadata.py
@@ -6,7 +6,9 @@ from collections.abc import Generator
 from configparser import ConfigParser
 from dataclasses import dataclass, field
 from pathlib import Path
-from typing import TYPE_CHECKING, Any, TypedDict
+from typing import TYPE_CHECKING, Any
+
+from typing_extensions import Required, TypedDict
 
 if TYPE_CHECKING:
     from configparser import SectionProxy
@@ -170,11 +172,11 @@ class Metadata:
         raise ValueError(f"Directory '{root}' does not exist.")
 
 
-class MetadataDict(TypedDict):
+class MetadataDict(TypedDict, total=False, closed=True):
     """TypedDict for app metadata."""
 
-    title: str
-    description: str
+    title: Required[str]
+    description: Required[str]
     authors: str | None
     state: str | None
     documentation_url: str | None
diff --git a/pyproject.toml b/pyproject.toml
index 74af772..c359c48 100644
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -28,6 +28,7 @@ dependencies = [
     "toml~=0.10",
     "traitlets~=5.0",
     "watchdog~=6.0",
+    "typing-extensions~=4.13",
 ]
 description = "Implements core functions of AiiDAlab"
 license = {file = 'LICENSE'}
```

danielhollas (approved):
Nice, this is better, thanks! I have two more suggestions, feel free to merge afterwards.

edan-bainglass:
@danielhollas I'll have a look, but FYI, this is massively wasting my time 😅

aiidalab/app.py
edan-bainglass: I had it this way first, but mypy complained
```
Incompatible return value type (got "str | None", expected "str")  [return-value]
```

aiidalab/metadata.py
edan-bainglass: You know too much about this stuff

aiidalab/app.py
danielhollas: Ah, you're right, because the type of `authors` field is `str | None`. Carry on

danielhollas (approved):
Cool, thank you

danielhollas:
@edan-bainglass I've pushed a commit with a mypy update, the old mypy version didn't support the `closed` argument. Sorry about that. You can merge once GitHub gets its shit together.

edan-bainglass:
> Right, that's likely a better solution, but also there are other unrelated `dict[str, Any]` types that could be improved. I propose we deal with it all at once in another PR.

There was a reason why I asked to handle the typing inaccuracies in another PR 😅 It is now obvious that the PR has gone far from its original goal - fixing a general `str`-casting via f-string that is no longer a correct assumption due to the new citations field. However, this opened a can of type-worms that has obfuscated the bug fix and delayed its release.

@danielhollas has requested/pushed changes towards a resolution. That's great! However, due to time constraints, I defer the completion of this PR to @danielhollas. Happy to review once satisfactory 🙂

danielhollas:
> There was a reason why I asked to handle the typing inaccuracies in another PR 😅 It is now obvious that the PR has gone far from its original goal - fixing a general str-casting via f-string that is no longer a correct assumption due to the new citations field. However, this opened a can of type-worms that has obfuscated the bug fix and delayed its release.

I've reverted the latest changes that went beyond the original scope. I'll see if I can add a small test for the bug being fixed here, and will merge after.

danielhollas:
Sorry for the crossed wires on this PR. I'll be more careful when wording comments to prevent misunderstandings. 

Thanks for the improved types, I'll open an issue for a follow-up tomorrow. I think we need to spend a bit more time on testing here so that we don't go around in circles on this feature.

PR #527, opened 2026-09-09
Use `packaging.version.Version` when comparing versions

edan-bainglass:
This PR resolves a mismatch between registered release format 
(tag format, e.g., v26.06.11) and app metadata format (e.g., 26.6.11)
by leveraging `packaging.version.Version`, which sees the two as
equivalent. This solves the false negatives reported for AiiDAlab apps
leading to wrong "App incompatible" messaging.

### DIFF
--- a/aiidalab/app.py
+++ b/aiidalab/app.py
@@ -27,7 +27,7 @@ from uuid import uuid4
 import requests
 import traitlets
 from dulwich.errors import NotGitRepository
-from packaging.version import Version
+from packaging.version import InvalidVersion, Version
 from watchdog.events import (
     EVENT_TYPE_CLOSED_NO_WRITE,
     EVENT_TYPE_OPENED,
@@ -287,7 +287,7 @@ class _AiidaLabApp:
             if (
                 available_versions
                 and isinstance(installed_version, str)
-                and Version(installed_version) != Version(available_versions[0])
+                and not self._version_equal(installed_version, available_versions[0])
             ):
                 return AppRemoteUpdateStatus.UPDATE_AVAILABLE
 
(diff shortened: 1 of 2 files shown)


aiidalab/app.py
```diff
@@ -283,7 +284,11 @@ def remote_update_status(self, prereleases: bool = False) -> AppRemoteUpdateStat
 
             # Check whether the locally installed version is the latest release.
             available_versions = list(self.available_versions(prereleases=prereleases))
-            if available_versions and installed_version != available_versions[0]:
+            if (
+                available_versions
+                and isinstance(installed_version, str)
```
yakutovicha, line 289:
```suggestion
```
This is probably too much.

yakutovicha (approved):
small nitpick, otherwise good to go 👍

danielhollas:
Thanks!

Given the failing tests we will have to be more careful here, because the version strings are derived from git tags which can be anything so parsing them with packaging.Version can throw an error. 

Would be great to have an additional test here that demonstrates the original problem.

aiidalab/app.py
edan-bainglass: It narrows the type down, to better support static type checkers :)

edan-bainglass:
> Given the failing tests we will have to be more careful here, because the version strings are derived from git tags which can be anything so parsing them with packaging.Version can throw an error.

Thanks @danielhollas. I now try/except comparison via `Version`, falling back on string comparison.
 
> Would be great to have an additional test here that demonstrates the original problem.

I added a simple test of the `_versions_equal` helper. Perhaps a more robust test can be added against the metadata. I'm skipping this for now, as I am not fully familiar with the testing suite in this repo. The original issue is documented in the description 🙂

aiidalab/app.py
danielhollas: `installed_version = self.installed_version()`

and `installed_version()` returns `str | AppVersion`. So I am not sure if this check is not papering over an actual issue that the type-checker points out. Do we need to specially handle the case where `installed_version` is an instance of AppVersion?

edan-bainglass: Indeed, this came up. I had to handle it explicitly:
```python
    def _versions_equal(self, left: str | AppVersion, right: str | AppVersion) -> bool:
        """Check if two versions are equal, considering PEP 440."""
        if not isinstance(left, str) or not isinstance(right, str):
            return left is right

        try:
            return Version(left) == Version(right)
        except InvalidVersion:
            return left == right
```

edan-bainglass: I think we can now remove the check and allow both `str` and `AppVersion` into `_versions_equal`, as both types are handled explicitly. Will do if we agree.

edan-bainglass:
@danielhollas may we proceed with this? 🙂

danielhollas (approved):
I don't have time today to do careful review. Happy to go ahead if this has been tested manually and things are working. 

Thanks for fixing this @edan-bainglass

